PR #81, opened 2012-05-14
Make it possible to use YAML to serialize and deserialize a mock object

floehopper:
Previously this was raising an exception because Mock#respond_to? was
being called before the instance of Mock has been re-hydrated i.e. none
of its instance variables had been set :-

```
NoMethodError: undefined method `matches_method?' for nil:NilClass
```

from gems/mocha-0.10.5/lib/mocha/mock.rb:185:in `respond_to?'

Note that I would not recommend using stubs in this way, but I decided
it was a shame that an exception was raised unnecessarily.

Note also that any expectations set on the original instance of Mock
will not be satisfied by invoking method on the re-hydrated instance of
Mock, because the two instances are completely different. This might be
solved by de-registering the Mock object when an instance with the same
"identitifer" is registered, but then I start wondering whether the
whole thing is a silly idea and we'd be better off raising an exception
if anyone attempts to serialize an instance of Mock.

### DIFF
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -209,6 +209,7 @@ module Mocha
 
     # @private
     def respond_to?(symbol, include_private = false)
+      return false if instance_variables.empty?
       if @responder then
         if @responder.method(:respond_to?).arity > 1
           @responder.respond_to?(symbol, include_private)
--- a/test/acceptance/stub_serialization_test.rb
+++ b/test/acceptance/stub_serialization_test.rb
@@ -1,0 +1,27 @@
+require File.expand_path('../acceptance_test_helper', __FILE__)
+require 'mocha'
+require 'yaml'
+
+class StubSerializationTest < Test::Unit::TestCase
+
+  include AcceptanceTest
+
+  def setup
+    setup_acceptance_test
+  end
+
+  def teardown
+    teardown_acceptance_test
+  end
+
+  def test_should_serialize_and_unserialize_successfully
+    test_result = run_as_test do
+      before = mock('before')
+      before.stubs(:bar).returns('bar')
+      dump = YAML.dump(before)
+      after = YAML.load(dump)
+      assert_equal 'bar', after.bar
+    end
+    assert_passed(test_result)
+  end
+end


floehopper:
> There is now a way to make the psych engine fail-fast - see https://github.com/tenderlove/psych/issues/112.

@chrisroos: I'll have a look into implementing the fail-fast behaviour and getting this merged.

floehopper:
I've had a look at implementing the fail-fast behaviour, but it would mean implementing `Mocha::Mock#to_yaml` (for Syck) and `#encode_with` (for Psych) methods to raise an exception. It would also mean adding a condition to `Mocha::Mock#respond_to?` to ensure it returns `true` for these methods. I'm now less convinced this is a good solution.

@chrisroos: Do you have any thoughts?

floehopper:
I've rebased this branch against master and fixed a problem caused by the change from `Test::Unit::TestCase" -> "Mocha::TestCase", and pushed it up as a [new branch](https://github.com/freerange/mocha/compare/fix-yaml-serialisation-for-stub-objects) to see whether the Travis CI build passes.

floehopper:
So it turns out that the [Travis CI build passed](https://travis-ci.org/freerange/mocha/builds/172261883) for the rebased version of this PR.

However, the more I've investigated this the less sure I am about the best solution. It could be argued that the commit in this PR does at least make things better in that no error will occur if you serialize a stub object to YAML and then re-load it. In fact any _stubbed_ methods will work as expected on the re-loaded instance.

However, as the commit note makes clear, expected methods are more of a problem. Currently expectations are stored against the mock object instance. But when an object is serialized and de-serialized, a new instance is created. Thus the expectations now exist in both the original instance and the new instance. It's likely that this will lead to confusing behaviour. A possible solution might be for the two instances to _share_ the same expectations. However, I'm not sure that's even possible and/or it's likely to require substantial re-working of the code.

I've found the following code in RSpec which might be relevant:

* https://github.com/rspec/rspec-mocks/blob/master/spec/rspec/mocks/serialization_spec.rb
* https://github.com/rspec/rspec-mocks/blob/master/spec/rspec/mocks/marshal_extension_spec.rb
* https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/marshal_extension.rb

And this documentation:

* http://www.rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FConfiguration%3Apatch_marshal_to_support_partial_doubles%3D

@chrisroos: I'd like to pair on this to try to come up with a sensible approach.

floehopper:
@chrisroos_gfr:

Thanks for chatting about this yesterday and for pairing with me on it. I did a bit more work on it yesterday evening and I ran into some more problems...

I got the tests we were trying to write to work by using `const_set` & `remove_const` (see [this gist][1]). However, the class method & any instance method tests work without any changes to Mocha, because by default when a class is serialised, none of its "state" is serialised (e.g. class variables, class instance variables). Thus I'm not sure these tests are actually useful.

I then tried running the whole test suite with our `encode_with` change in place. However, I saw a load of warnings due to instance variables in the de-serialised mock object not being initialised. We hadn't seen these previously because warnings are not enabled by default.

It also then dawned on me that our tests don't really check much of the behaviour of the de-serialised mock object and in fact some of it is definitely broken, because the instance variables are not set, e.g. unexpected invocations.

I started trying to set the instance variables to something sensible, but ran into problems, e.g. with the `@mockery` instance variable. I'm now wondering whether the aim of having the de-serialised mock object effectively be a new instance of a mock object with the same name is a sensible one.

I'm now leaning back towards failing fast if someone tries to serialize a mock object. Given that our `encode_with` solution involves adding methods to the `Mock` class anyway, it doesn't feel so bad to add such methods and raise a `TypeError`.

Let me know what you think - either in here or when we next meet up.

[1]: https://gist.github.com/floehopper/2e3915ef4ef64ed2d487c30f2d350b42

floehopper:
Closing in favour of #318, i.e. the fail fast approach.

PR #208, opened 2014-09-26
Require at least ruby -v 1.8.7.

knappe:


chrisroos:
I've done a bit of investigation:

It doesn't look as though Ruby versions earlier than 1.8.7 are available on Travis although I guess it might be possible to manually install an earlier version.

I used rbenv to try to install ruby-1.8.5-p113 on my laptop. It failed.

I used rbenv to install ruby-1.8.6-p110 on my laptop but gave up trying to get the Mocha tests to run when I failed to install Bundler.

floehopper:
@chrisroos:

> I used rbenv to try to install ruby-1.8.5-p113 on my laptop. It failed.
>
> I used rbenv to install ruby-1.8.6-p110 on my laptop but gave up trying to get the Mocha tests to run when I failed to install Bundler.

I had a go at using rbenv to install ruby-p420 last night, but I ran into "abort trap 4" or something similar when trying to install bundler. I'm planning to have a go at installing v1.8.6 on an older laptop either tonight or tomorrow. If that fails, maybe we should accept this PR.

floehopper:
I've had a couple of attempts at installing Ruby 1.8.6, firstly on an old laptop and secondly on a Virtualbox instance running Ubuntu, but without success.

Given that we can't verify that Mocha works on Ruby 1.8.6, perhaps it's OK to restrict it to Ruby 1.8.7+.

@chrisroos: What do you think?

chrisroos:
Restricting it to 1.8.7+ seems reasonable to me, @floehopper. I'm sure we'll hear if people really are still using it with older versions of Ruby.

Are you happy for me to apply the changes in this PR?

knappe:
For what it is worth, 1.8.7 was retired over three years ago.  https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/  It also hasn't seen security patches since June 2014 https://www.ruby-lang.org/en/news/2013/12/17/maintenance-of-1-8-7-and-1-9-2/.  I'd really think we don't want to encourage the use of 1.8.7.

floehopper:
@knappe: I understand that, but I don't think it's Mocha's job to force people to upgrade when Mocha will work quite happily. The last thing I want to do is to cause problems to people who are already using Mocha in a project which is using an earlier version of Ruby.

floehopper:
@chrisroos:

> Are you happy for me to apply the changes in this PR?

Yes. Although maybe this restriction ought to be mentioned in the README too...?

floehopper:
I've just rebased this against master and pushed it up in the [require-at-least-ruby-1.8.7 branch](https://github.com/freerange/mocha/compare/require-at-least-ruby-1.8.7). If the Travis CI build passes, I plan to merge the PR.

floehopper:
Manually merged in 3e20be8e0d8289b47c2194c8fe0951d8e2a13b93. Closing.

PR #217, opened 2015-03-02
Allow `includes` matcher to take matcher arguments.

lazyatom:
We should be able to created a matcher of the form:

    object.expects(:foo).with(includes(has_entry(:key, 'value')))

and successfully match this against a method call like:

    object.foo([:a, :b, {:key => value}, :c])

Hopefully this makes sense; let me know if you've already considered functionality like this but decided against it.

### DIFF
--- a/lib/mocha/parameter_matchers/includes.rb
+++ b/lib/mocha/parameter_matchers/includes.rb
@@ -38,9 +38,8 @@ module Mocha
       def matches?(available_parameters)
         parameter = available_parameters.shift
         return false unless parameter.respond_to?(:include?)
-
         if @items.size == 1
-          return parameter.include?(@items.first)
+          return parameter.any? { |p| @items.first.to_matcher.matches?([p]) }
         else
           includes_matchers = @items.map { |item| Includes.new(item) }
           AllOf.new(*includes_matchers).matches?([parameter])
--- a/lib/mocha/parameter_matchers/includes.rb
+++ b/lib/mocha/parameter_matchers/includes.rb
@@ -23,13 +23,13 @@ module Mocha
     #   object.method_1(['foo', 'baz'])
     #   # error raised, because ['foo', 'baz'] does not include 'bar'.
     #
-    # @example Actual parameter includes item which matches nested matcher
+    # @example Actual parameter includes item which matches nested matcher.
     #   object = mock()
     #   object.expects(:method_1).with(includes(has_key(:key)))
     #   object.method_1(['foo', 'bar', {:key => 'baz'}])
     #   # no error raised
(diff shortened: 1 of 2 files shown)


floehopper:
@lazyatom: Thanks for the PR. It does feel as if there's some behaviour missing/inconsistent in this area. I'll have a bit of a ponder.

lazyatom:
I've made changes based on your comments, thanks @floehopper. Let me know what you think.

lazyatom:
New behaviour to more explicitly handle `String` and `Hash` arguments is now present.

Essentially, if the object responds to `any?` then we use that to match the values, and the first yielded value is used for comparison with the given value or matcher. If the object does *not* respond to `any?`, then the original behaviour of directly calling `includes?` on that object is used.

floehopper:
@lazyatom: Thanks for doing all this. I've just noticed that both `Range` and `Module` implement a `#include?` method. I'm guessing your new implementation will work with both of these. Do you think that's right?

lazyatom:
`Range` should work fine as it supports `any?` and yields each value in turn in the same way that `Array` does.

`Module` should also work because it does *not* respond to `any?` and so the original behaviour of `parameter.include?` will get called. Or, in other words, using this matcher with a module is no more tricky before or after this pull request ;)

I can't think of an existing matcher that you might want to nest and use against a module argument for a method (the closest I could think of was `kind_of` or `instance_of` but they both don't make sense when including modules in other modules (or classes); if module/class B includes module A, module B is still not a 'kind of' A, i.e. `B.kind_of?(A) => false`), so I *think* it might be fine. Future new matchers might make this an issue to revisit perhaps?

floehopper:
@lazyatom:

Thanks for all the work you've done on this.

I noticed that there's a [failing test in Ruby 1.8.7](https://travis-ci.org/freerange/mocha/jobs/53116525) which seems to be because (unlike in other versions of Ruby) `String` *does* respond to `any?`. The implementation of `String#any?` seems to yield the whole `String` to the block in one go which then doesn't give the expected matcher behaviour.

Do you think this PR is otherwise ready to merge?

lazyatom:
Other than that case, and if you're happy with the documentation, I think it's ready...

In order to fix that case, I cannot think of an approach other than explicitly checking if the parameter is a `String`. I'll implement that and we'll see how it looks.

lazyatom:
I think this is good to go. Happily leaving it in your hands for now :)

floehopper:
@lazyatom: Thanks. Sorry for the delay - I've been away. I've taken the liberty of squashing some of the commits, but I've just merged your changes. Closing.

PR #227, opened 2015-04-30
Pass result of yield to caller

steverice:
When using `#yields` or `#multiple_yields`, the expectation will only return a value if it is explicitly specified with a `#returns` call.
The default behavior of Ruby blocks is to pass the result of the block back to the caller.

This mimics that behavior with expectations that yield. The result of the yielded block is the return value of the expectation unless a return value was explicitly specified. This allows stubbing out the logic of a method that ultimately yields to a given block without breaking its caller's expectation of the return value it receives, vs. a `#yield` stub implicitly stubbing out the behavior of everything it yields to as well.

In the case of `#multiple_yields`, the result from the final yield is returned, as if the stubbed method was executing:
```ruby
yield(first_parameter_group)
yield(second_parameter_group)
…
yield(final_parameter_group)
```

Fixes https://floehopper.lighthouseapp.com/projects/22289/tickets/14-8687-blocks-return-value-is-dropped-on-stubbed-yielding-methods

In the referenced ticket, it was indicated that this behavior was not in line with the Mocha philosophy and this behavior is not desired. In any case, this PR serves as a proof of concept of how it could be done.

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -562,9 +562,10 @@ module Mocha
       @invocation_count += 1
       perform_side_effects()
       if block_given? then
-        @yield_parameters.next_invocation.each do |yield_parameters|
+        yield_result = @yield_parameters.next_invocation.each do |yield_parameters|
           yield(*yield_parameters)
         end
+        return yield_result if @return_values.values.empty?
       end
       @return_values.next
     end
--- a/lib/mocha/multiple_yields.rb
+++ b/lib/mocha/multiple_yields.rb
@@ -9,9 +9,9 @@ module Mocha
     end
 
     def each
-      @parameter_groups.each do |parameter_group|
+      @parameter_groups.map do |parameter_group|
         yield(parameter_group)
-      end
+      end.last
     end
 
   end
--- a/test/unit/expectation_test.rb
+++ b/test/unit/expectation_test.rb
@@ -147,6 +147,24 @@ class ExpectationTest < Mocha::TestCase
     assert_equal [[1, 2, 3], [4, 5], [6, 7]], yielded_parameters
   end
 
+  def test_should_return_yielded_value
+    expectation = new_expectation.yields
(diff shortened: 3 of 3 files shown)


floehopper:
@steverice: Thanks for the PR. I'm very busy at the moment, but I'll try to have a look as soon as I can.

steverice:
@floehopper any consideration for this PR?

floehopper:
@steverice: I'm sorry - I'm really busy at the moment. I'll try to have a look in the next 2-3 weeks. Given my response to the original lighthouse ticket, I wouldn't hold out a lot of hope that I'll accept this PR. So if there's anything you can add here to persuade me it's a good idea, that would be useful. In particular, do you have some concrete examples where you want to use this where the current implementation is lacking?

floehopper:
I'm sorry that I haven't found time to look at this any further. However, I note that you never got back to me with any concrete examples where you'd want to use this functionality. So I'm going to close this PR for now. Please feel free to re-open it if you have further input.

steverice:
@floehopper this one got away from me too.

Here's a simplified example of one place I'm using this: a locking library that performs a set of actions with a ZooKeeper lock:
```ruby
class ZooKeeperLock
  def self.with_lock(*args)
    # Ensure connection and establish lock
    …
    # Let the code run under the lock
    yield
    # Release the lock
    …
  end
end
```

In unit tests, I'm not looking to test the integration with ZooKeeper. However, all the code that calls the locking library is expecting a call to `ZooKeeperLock.with_lock { puts 'hello world' }` to yield to the given block. Since different callers are doing a number of different things with the yielded block — and it's part of the code under test — it's not really feasible to set a number of expectations with fixed/known return values.

Without this patch, there's no way I'm aware of to remove ZooKeeper from the unit tests using Mocha. With the patch, I call `ZooKeeperLock.stubs(:with_lock).yields` and the code works as expected.

FWIW, we have been using this patch in a fairly substantial Rails codebase since I opened the PR and have not experienced any ill side effects.

Issue #224, opened 2015-03-25
Consider providing a way to define jMock-style custom actions


floehopper:
See http://www.jmock.org/custom-actions.html.

nitishr:
I'm not sure I understand custom actions properly. With my limited understanding, I suspect #379 may provide the described capability.

If not, I'd love to help implement this feature if you could explain to me the desired behavior. A ruby/mocha hypothetical example would be great.

floehopper:
I've changed the title of this issue, because I want to use it to roll up a number of closely related issues and have the discussion in one place.

floehopper:
> I'm not sure I understand custom actions properly. With my limited understanding, I suspect #379 may provide the described capability.
> 
> If not, I'd love to help implement this feature if you could explain to me the desired behavior. A ruby/mocha hypothetical example would be great.

@nitishr Thanks for asking about this.

You're right that #213, #230, #231, #317 and your latest #379 are all closely related to this issue. I'm going to close all those and continue the discussion here.

floehopper:
My hesitation about implementing this seemingly innocuous extra  functionality is that it seems to violate the idea of a stub always returning "canned" data and thus risk the test becoming very complicated and hard to follow.

Passing a block or proc to be invoked by the stub means allowing any amount of _arbitrary_ code to be executed by the stub including the possibility of mutating state _outside_ the stub or even _global_ state. This feels inherently dangerous and I can imagine people getting in a real mess with it.

I think that the execution flow of a test should go from the test (T) to the object under test (O) and then into one or more stubbed methods (S) and then back out to the object under test and ultimately the back to the test again:

```
T -> O
     O -> S
     O <- S
T <- O
```

However, if you allow an arbitrary block to be invoked by the stub, you're likely to end up with something more like:

```
T -> O
     O -> S
T <------ S
T ------> S
     O <- S
T <- O
```

And I think this would quickly become hard to follow.

[...]

PR #231, opened 2015-07-22
Allow generation of return values by block with invocation arguments

mattbornski:


mattbornski:
@floehopper for your consideration

kpatrick:
I want this too.

chrisroos:
Hi @mattbornski and @kpatrick. Apologies it's taken such a long time to respond to this PR!

Can you help me understand the use case for this functionality? I haven't given it any real thought (and may well be misunderstanding the purpose of this change) but it seems slightly odd that I'd want to define some behaviour in a stubbed method.

mattbornski:
@chrisroos I'm looking to see if I can find the specific case where I needed this any longer, but off the top of my head I believe one of the major reasons is for meta-programming tests.  For instance a function wherein the response includes the input somehow; if this gets called 12 times in a suite of tests I might prefer to set the expectation once with a block instead of each time with a static value.

kpatrick:
I've switched to a non-ruby project recently, so this doesn't impact me at this time.

chrisroos:
Thanks for the responses, @kpatrick and @mattbornski.

@mattbornski: Did you have any luck finding the specific case where you needed this behaviour? I'd really like to understand more about the problem this is trying to solve and seeing some real world code might help me do that.

mattbornski:
The code in question is in a private repo, but imagine you have some static data (`foo.name`, `foo.features`) and some dynamically generated data (`foo.uuid`).  In your test you can of course lock down a lot of data, but you might not want/be able to effectively lock down some of the dynamically generated data (perhaps it comes from a DB in an integration test and it's simply not worth the effort and/or so invasive as to undermine the entire value of the test to force it to be static).

Now imagine you have a function in which the response is dynamic based on the input.  Because the function calls external services or is not core to the test, you don't want to actually invoke it in tests but just simulate the response.  One of the more effective ways to stub this function is to use Mocha to generate a return value based on the input value.

```def transmogrify(obj)
  unless obj.features.include?(:can_transmogrify)
    return TransmogrificationResult(:error, "object #{obj.uuid} does not support transmogrification: features are #{obj.features}")
  end
  return TransmogrificationResult(:success, new CoreObject(UUID.new(), {features: features + ['transmogrified'], root_object: obj.uuid})
end
```

I definitely understand that the example is a little simplistic but the use case of wanting to mock something dynamically based on the input is very real.

chrisroos:
Thanks for following up with the example, @mattbornski. Unfortunately, I'm still struggling to get my head around when I'd use this.

Using your example code, I could imagine having tests like:

```ruby
def test_transmogrify_when_object_can_transmogrify
  transmogrifier = Transmogrifier.new
  object = mock('foo', 
    features: [:can_transmogrify],
    uuid: 'object-uuid'
  )
  result = transmogrifier.transmogrify(object)
  # assert success of transmogrification
end

def test_transmogrify_when_object_cannot_transmogrify
  transmogrifier = Transmogrifier.new
  object = mock('foo', 
    features: [:cannot_transmogrify],
    uuid: 'object-uuid'
  )
  result = transmogrifier.transmogrify(object)
  # assert failure of transmogrification
end
```

Can you illustrate how you'd use the functionality added in this PR to write similar tests? Maybe that'll help me understand what I'm currently missing.

mattbornski:
Perhaps in a bigger test, such as an integration test, where the complexity
is harder to factor out?

In any case I'm using the branch of the PR without issues so if you
fundamentally don't see the need, my problems at least are already handled.

chrisroos:
Hi @mattbornski. I appreciate the effort you've put into this PR and in trying to help me understand the problem it's solving. Unfortunately, I'm still unable to see when this would be useful. Given that you're currently using your own branch and I can't yet justify merging this to master I'm going to close this PR. I'm certainly still happy to be persuaded that this is useful so feel free to reopen it if you want to the continue the discussion.

floehopper:
A relevant discussion continues in #224.

PR #240, opened 2016-05-20
adds a matches_fn parameter matcher

bkoski:
This adds a `matches_fn()` parameter matcher where you can define an arbitrary lambda to check a param.

This is useful for logic not covered by other matchers (for example, if a JSON string needs to be parsed, or if there is some more complex match logic).


bkoski:
I'm not sure how to fix the build errors for ruby < 2.0.  It seems that it's some kind of dependency issue that isn't showing up on the master build because the last build was 8 months ago?  @floehopper any advice?

floehopper:
@bkoski:

Thanks for the pull request. However, I believe this functionality is already available by passing a block to [Mocha::Expectation#with][1] - see the second example in the documentation.

In practice I often wrap the block up into a local variable or private method in order to make the code more readable e.g. this [a_sniper_that_is][2] method.

Thanks for the heads-up regarding the Travis CI build - I've created a [separate issue](https://github.com/freerange/mocha/issues/241) to track that.

Since I believe the functionality is already available, I'm going to close this PR. Please re-open if you disagree.

[1]: http://gofreerange.com/mocha/docs/Mocha/Expectation.html#with-instance_method
[2]: https://github.com/freerange/goos-ruby/blob/f50bfb91a631fbf602b834093608f2aad59cbd26/test/unit/auction_sniper_test.rb#L21-L24

PR #202, opened 2014-06-06
Fix alias method from included module on 2.1 and 2.0.0-p451

grosser:
@floehopper

PR #243, opened 2016-06-21
alias method fix

grosser:
replacement PR for https://github.com/freerange/mocha/pull/202

since I just ran into this again and the original has `Unknown repository` ...


grosser:
wrong origin, closing

Issue #241, opened 2016-05-22
Investigate failing Travis CI build for Ruby < v2.0


floehopper:
See #240.

PR #244, opened 2016-06-21
Fix alias method from included module on 2.1

grosser:
replacement PR for #202

since I just ran into this again and the original has `Unknown repository` ...

### DIFF
--- a/lib/mocha/any_instance_method.rb
+++ b/lib/mocha/any_instance_method.rb
@@ -69,7 +69,7 @@ module Mocha
         !(Class === mod)
       end
 
-      if possible_prepended_modules.any?
+      if possible_prepended_modules.any? || aliased?
         @definition_target = PrependedModule.new
         stubbee.__send__ :prepend, @definition_target
       end
--- a/lib/mocha/any_instance_method.rb
+++ b/lib/mocha/any_instance_method.rb
@@ -69,7 +69,7 @@ module Mocha
         !(Class === mod)
       end
 
-      if possible_prepended_modules.any? || aliased?
+      if possible_prepended_modules.any?
         @definition_target = PrependedModule.new
         stubbee.__send__ :prepend, @definition_target
       end
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -112,7 +112,7 @@ module Mocha
         !(Class === mod)
       end
 
-      if possible_prepended_modules.any?
+      if possible_prepended_modules.any? || aliased?
         @definition_target = PrependedModule.new
         stubbee.__metaclass__.__send__ :prepend, @definition_target
       end
@@ -120,6 +120,10 @@ module Mocha
 
     def definition_target
       @definition_target ||= stubbee.__metaclass__
+    end
+
+    def aliased?
+      @original_method.original_name != @original_method.name
     end
 
   end
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -6,6 +6,8 @@ module Mocha
 
     PrependedModule = Class.new(Module)
 
+    RUBY_21_PLUS = (RUBY_VERSION >= "2.1.0")
+
     attr_reader :stubbee, :method
 
     def initialize(stubbee, method)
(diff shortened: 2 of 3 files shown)


floehopper:
@grosser: Thanks for (re-)submitting this PR. I'm investigating the CI build failures.

floehopper:
Having now [fixed the Travis CI build on master][1], I've created my own version of the PR branch and pushed it to GitHub. This has kicked off a [new build][2] for the branch.

[1]: https://github.com/freerange/mocha/issues/241
[2]: https://travis-ci.org/freerange/mocha/builds/140066218

floehopper:
I'm seeing [a single test failure][1] when running under Ruby v1.8.7 and using the built-in version of test-unit.

[1]: https://travis-ci.org/freerange/mocha/jobs/140066219#L233

floehopper:
At the risk of repeating what I said in [a comment on the original PR][1]:

* Can you explain how the scenario in your acceptance test is different from this other [existing acceptance test][2]?
* Does method aliasing somehow work differently in Ruby >= v2.0?
* Does method aliasing work differently when done via extending a module?
* Can you explain why Mocha currently does the wrong thing in the scenario you're trying to fix?
* Is the acceptance test the simplest scenario which reproduces the problem? - it seems quite convoluted.
* Can the acceptance test be modified to cover the any_instance & instance method equivalents of this scenario?

I have to admit I'm pretty confused.

[1]: https://github.com/freerange/mocha/pull/202#issuecomment-45483068
[2]: https://github.com/freerange/mocha/blob/91288948057473d58377a9ab642031fd6eddc9c7/test/acceptance/stub_instance_method_defined_on_class_and_aliased_test.rb#L16..L32

grosser:
I'm not really sure either ... I'll have to dig back into this patch :D

grosser:
managed to simplify the test case 1 step ... but it only happens when a inherited base method is aliased and then the aliased method is stubbed

was not able to reproduce for instance methods ... so this possibly was added by accident ... removed ...

the difference to the existing test is that it is overwriting a method on the same class and not an inherited method 

... I did not dive deeper into why the logic in mocha is wrong / how extending a class is different / how it is different in ruby 2.1 ...

grosser:
updated to fix ruby 1.8 test case ... good enough ?

chrisroos:
We haven't forgotten about this, @grosser.

@floehopper and I spent quite a while investigating the problem earlier this week and think/hope we can come up with a solution that works for Ruby 1.8.7 and up.

grosser:
❤️

chrisroos:
Hi @grosser. @floehopper and I have just pushed some changes to master. Would you be able to see whether the master branch version of Mocha fixes your problem? If it does we'll aim to release a new version shortly.

grosser:
fixes the issue ... but it broke something else:

```Ruby
class A
  def method_missing(*args)
    1
  end
end

a = A.new
a.expects(:warn) # a private method inherited from Kernel ...
a.warn

- expected exactly once, not yet invoked: #<A:0x7fcaa9eed3a8>.warn(any_parameters)
```

... we got the worst code, just so you can find all edge-cases ;)

FYI using `send :warn` makes it work properly ...

chrisroos:
Thanks for the update, @grosser. We'll investigate the new problem.

chrisroos:
@grosser:

I've created issue #260 to capture the most recent problem you've described. Can you take a look and see whether you think I've understood correctly, please?

Ignoring the new issue for a moment, are you happy for us to close this PR given the fix we've applied to master?

grosser:
Yes, this issue should be gone ... and there is a test in case is comes
back :)

grosser:
FYI unrelated but good change: some stub or expect results used to be dupped strings (not sure exactly what edge-case caused that), but now they are no longer dupped :)

grosser:
all our project builds are green now ... want to release ?

chrisroos:
That's great! We want to finish #262 before releasing. I hope to get to
that within the next week.

Thanks for your help, @grosser.

floehopper:
@grosser: [Mocha v1.2.0](https://rubygems.org/gems/mocha/versions/1.2.0) has just been released incorporating these changes. Thanks for your help & patience!

PR #248, opened 2016-06-27
dry-up

grosser:
@floehopper 

 - reuse ruby version check outcome instead of re-doing it every time
 - reuse method visibility instead of re-doing it with slightly different logic which could lead to subtle bugs

### DIFF
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -4,6 +4,9 @@ module Mocha
 
   class ClassMethod
 
+    RUBY_18 = (RUBY_VERSION < '1.9')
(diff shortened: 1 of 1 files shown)


floehopper:
@grosser: Thanks for the PR. Sorry for the slow response. I've incorporated the ideas in this PR into master in [this series of commits][1]. Closing.

[1]: https://github.com/freerange/mocha/compare/d2808991ea5bcda3095c572c143b737ea90434d5...c53fdaa00687961916ec19d1e6b427a93a595cc3

chrisroos:
We've realised we need to make these same changes to `AnyInstanceMethod#hide_original_method` (see PR #262 for the work in progress). We want to do that before releasing the next version of Mocha.

PR #254, opened 2016-08-24
Fix typo in docs for equals

alexcoco:
@floehopper 

The docs say that an exception is raised because `method_1` is not called with an object that equals `3` but the expectation is that the object equals `2` (`with(equals(2))`).


floehopper:
@alexcoco: Thanks!

PR #263, opened 2016-09-08
Add Ruby 2.3 to Travis config

chrisroos:
It makes sense to test against Ruby 2.3 given that it's the current
stable version.


floehopper:
Looks good to me! 👍

floehopper:
Rebasing against master and force-pushing in preparation for merging.

PR #267, opened 2016-09-29
Ensure all tests run individually

### LINKED ISSUE #251: Some tests in ClassMethodTest fail when it is run on its own
```
$ bundle exec ruby -Itest test/unit/class_method_test.rb
Run options: --seed 32684

# Running:

...F....E.E....E...E

Finished in 0.002676s, 7473.2355 runs/s, 5604.9267 assertions/s.

  1) Failure:
ClassMethodTest#test_should_hide_original_method [test/unit/class_method_test.rb:17]:
Expected: false
  Actual: true


  2) Error:
ClassMethodTest#test_should_call_remove_new_method:
NoMethodError: undefined method `reset_mocha' for #<Class:0x007fe458a011f0>
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:37:in `reset_mocha'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:28:in `unstub'
    test/unit/class_method_test.rb:125:in `test_should_call_remove_new_method'


  3) Error:
ClassMethodTest#test_should_restore_original_method_accepting_a_block_parameter:
NameError: undefined method `method_x' for class `Class'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `public'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `call'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `restore_original_method'
    test/unit/class_method_test.rb:78:in `test_should_restore_original_method_accepting_a_block_parameter'


  4) Error:
ClassMethodTest#test_should_call_restore_original_method:
NoMethodError: undefined method `reset_mocha' for #<Class:0x007fe4589f0440>
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:37:in `reset_mocha'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:28:in `unstub'
    test/unit/class_method_test.rb:138:in `test_should_call_restore_original_method'


  5) Error:
ClassMethodTest#test_should_restore_original_method:
NameError: undefined method `method_x' for class `Class'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `public'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `call'
    /Users/jamesmead/Code/freerange/mocha/lib/mocha/class_method.rb:82:in `restore_original_method'
    test/unit/class_method_test.rb:65:in `test_should_restore_original_method'

20 runs, 15 assertions, 1 failures, 4 errors, 0 skips
```

chrisroos:
I'd noticed that running some tests individually (i.e. outside of `rake test`) would result in failures.

I created a script to run each test file individually which allowed me to find and fix problems in the following two tests:

* test/unit/any_instance_test.rb
* test/unit/class_method_test.rb

I believe this fixes #251.

### DIFF
--- a/test/unit/any_instance_method_test.rb
+++ b/test/unit/any_instance_method_test.rb
@@ -76,6 +76,7 @@ end
     klass.define_instance_method(:any_instance) { any_instance }
     method = AnyInstanceMethod.new(klass, :method_x)
     method.replace_instance_method(:restore_original_method) { }
+    method.replace_instance_method(:reset_mocha) { }
     method.define_instance_accessor(:remove_called)
     method.replace_instance_method(:remove_new_method) { self.remove_called = true }
 
@@ -92,6 +93,7 @@ end
     klass.define_instance_method(:any_instance) { any_instance }
     method = AnyInstanceMethod.new(klass, :method_x)
     method.replace_instance_method(:remove_new_method) { }
+    method.replace_instance_method(:reset_mocha) { }
     method.define_instance_accessor(:restore_called)
     method.replace_instance_method(:restore_original_method) { self.restore_called = true }
 
--- a/test/unit/class_method_test.rb
+++ b/test/unit/class_method_test.rb
@@ -121,6 +121,7 @@ end
     method = ClassMethod.new(klass, :method_x)
     mocha = build_mock
     klass.define_instance_method(:mocha) { mocha }
+    method.replace_instance_method(:reset_mocha) { }
     method.define_instance_accessor(:remove_called)
     method.replace_instance_method(:remove_new_method) { self.remove_called = true }
 
@@ -134,6 +135,7 @@ end
     mocha = build_mock
     klass.define_instance_method(:mocha) { mocha }
     method = ClassMethod.new(klass, :method_x)
+    method.replace_instance_method(:reset_mocha) { }
     method.define_instance_accessor(:restore_called)
     method.replace_instance_method(:restore_original_method) { self.restore_called = true }
 
--- a/test/unit/class_method_test.rb
+++ b/test/unit/class_method_test.rb
@@ -11,6 +11,7 @@ class ClassMethodTest < Mocha::TestCase
 unless RUBY_V2_PLUS
   def test_should_hide_original_method
     klass = Class.new { def self.method_x; end }
+    klass.__metaclass__.send(:alias_method, :_method, :method)
     method = ClassMethod.new(klass, :method_x)
 
     method.hide_original_method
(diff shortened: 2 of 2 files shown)


chrisroos:
@floehopper: Can you take a look at this when you have a moment please?

floehopper:
While I agree that the current behaviour is wrong and that your change fixes the problem, I'm not sure it's the right way to fix the problem. Requiring 'mocha/object_methods' only fixes the problem very indirectly - it only *defines* the `ObjectMethods` module - it's only because of the inclusion of 'test/method_definer' that this module is included into `Object`.

Requiring 'mocha/object_methods' also has the effect of requiring a load of other files. Given that this is intended to be a *unit* test, I don't think that's very desirable.

I suspect what's happened is that this test was originally "stubbing" (using 'test/method_definer' methods) various bits of `ClassMethod` in order to avoid dependencies on other files/classes/modules. As the implementation of `ClassMethod` has changed, the unit test hasn't kept up and is missing some "stubs".

Looking through the code, I think the re-opening of the `Mocha::ObjectMethods` module and its inclusion into `Object` is in a bit of a muddle. I've created a new issue, #268, to handle this.

chrisroos:
Thanks, @floehopper. I did wonder whether this was the correct solution.

I'll investigate stubbing the required methods instead of including `Mocha::ObjectMethods`.

chrisroos:
I've updated this branch so that I'm stubbing methods on `ClassMethod`, `AnyInstanceMethod` and the `stubbee` objects created in the tests. The commits contain a couple of different approaches as I'm not completely sure which way to go.

@floehopper: Do you have thoughts about:

1. Should I stub `ClassMethod#reset_mocha` or `stubbee.reset_mocha` in class_method_test and any_instance_method_test?

2. Should I use `define_instance_method(:_method)` or `__metaclass__.send(:alias_method, :_method)` in class_method_test?

floehopper:
@chrisroos:

> 1. Should I stub `ClassMethod#reset_mocha` or `stubbee.reset_mocha` in class_method_test and any_instance_method_test?

I think I would stub `ClassMethod#reset_mocha` unless an individual test is explicitly checking that `stubbee.reset_mocha` is being called; I'm not sure any of them currently are.

> 2. Should I use `define_instance_method(:_method)` or `__metaclass__.send(:alias_method, :_method)` in class_method_test?

I don't feel strongly either way. If I had to choose, I'd probably use  `__metaclass__.send(:alias_method, :_method)`, because it's closer to the implementation in `ObjectMethods`.

As I've previously mentioned, I'm not very happy with these tests. However, I'm happy that your changes don't make them any worse and it'd be great to have them passing when run on their own.

chrisroos:
Thanks for the feedback, @floehopper. I've now rewritten the commits in this branch and plan to merge if/when the tests have all passed.

Issue #286, opened 2016-11-26
Instance variables from the test case are not available inside API#stub block


ggaston095:
I think this is some kind of bug. I don't know why stub2 can't see stub1
```ruby
test "test1" do 
  @stub1 = stub do
    stubs(:method1).returns(0)
  end 
  stub2 = stub do 
    #error, @stub1 is nil
    stubs(:find_by_id).with(@stub1.method1).returns(@stub1)
  end
end
```

floehopper:
@ggaston095: I hope you don't mind, but I've tidied up the formatting of your code and switched on Ruby syntax highlighting to make things clearer.

floehopper:
@ggaston095: I've also renamed the issue to better explain the problem.

floehopper:
@ggaston095:

I've investigated this a little. The block passed to the `Mocha::API#stub` method is currently evaluated using `BasicObject#instance_eval`, so the block is evaluated in the context of the newly instantiated stub object and not in the context of the test case instance. This is why the instance variable from the outer context is not available inside the block.

I can see how this is confusing and I will have a think about what the best solution is. At the very least we should update the documentation to make it clearer what the current behaviour is.

Presumably the code you included in the description was example code and not real code. In case it's real code, you could solve your immediate problem by using a local variable (`stub1`) instead of the instance variable, `@stub1`.

ggaston095:
Thanks. I'll stick with the local variable for now :)

floehopper:
The relevant documentation has been improved and the mechanism itself has been deprecated in #290. These change will go out in the next release. Closing.

PR #290, opened 2016-12-01
Deprecate passing block to mock object constructor

floehopper:
This mechanism is confusing, because the block is evaluated in the
context of the newly instantiated block and not in the context of the
current test.

The mechanism is also of dubious value, because it's possible to use
Object#tap or define stubs/expectations with the mock object as the
explicit receiver without the code becoming much more verbose.

It's unfortunate that the deprecations make the tests more complicated,
but we will be able to simplify them again when this functionality is
removed in a later version.

See #286 for further details.

### DIFF
--- a/lib/mocha/api.rb
+++ b/lib/mocha/api.rb
@@ -27,6 +27,7 @@ module Mocha
     # @param [Hash] expected_methods_vs_return_values expected method name symbols as keys and corresponding return values as values - these expectations are setup as if {Mock#expects} were called multiple times.
     # @yield optional block to be evaluated in the context of the mock object instance, giving an alternative way to setup stubbed methods.
     # @yield note that the block is evaulated by calling Mock#instance_eval and so things like instance variables declared in the test will not be available within the block.
+    # @yield deprecated: use Object#tap or define stubs/expectations with an explicit receiver instead.
     # @return [Mock] a new mock object
     #
     # @overload def mock(name, &block)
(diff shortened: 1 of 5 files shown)


floehopper:
@chrisroos: I'd be grateful if you could review this at some point.

floehopper:
I've rebased against master and force-pushed in the hope of fixing the Travis CI build for this branch.

floehopper:
[Build is now fixed](https://travis-ci.org/freerange/mocha/builds/181305310).

floehopper:
I've now introduced an earlier commit which improves the documentation of the existing behaviour, before that behaviour is then deprecated in the subsequent commit.

@chrisroos: It'd be great if you could have a look at this when you have a chance. I don't think it should take you long. Let me know if you want help on how to generate the documentation...

chrisroos:
Out of interest, why have you used `DeprecationDisabler.disable_deprecations` (i.e. the module method) in mock_with_initializer_block_test.rb and `disable_deprecations` (i.e. the included instance method) in mockery_test.rb?

As an observation, I presume there aren't any tests for this now-deprecated behaviour for `stub` and `stub_everything`.

Aside from my question and observation this all looks good to me, @floehopper.

floehopper:
@chrisroos:

Thanks for reviewing.

> Out of interest, why have you used `DeprecationDisabler.disable_deprecations` (i.e. the module method) in mock_with_initializer_block_test.rb and `disable_deprecations` (i.e. the included instance method) in mockery_test.rb?

In the former, each test is instantiating a new instance of a test case (via `TestRunner#run_as_test) and the `DeprecationDisabler` module would need to be included into each instance. So I decided it was simpler to just use the module method in this case. Does that make sense?

> As an observation, I presume there aren't any tests for this now-deprecated behaviour for `stub` and `stub_everything`.

You are correct. I think I'm OK with this on the basis I know that the behaviour is implemented within `Mock#initialize`, i.e. I'm using a pragmatic white-box testing approach.

I'm planning to get this merged, so do shout if you have any more comments or questions.

PR #294, opened 2017-01-28
Use do/end instead of { } in stub_module_method_test

chrisroos:
This is consistent with the way we're defining modules and classes in
other similar tests. I also find it easier to read.


chrisroos:
I'm fairly confident this can just be merged but I don't have time to look at it right now so I'm opening this PR as a reminder that I need to do something with it.

floehopper:
@chrisroos: I'm going to make the same change across all tests and add them to this PR. I hope that's OK.

PR #295, opened 2017-01-28
Remove `ModuleMethod#method_exists?`

chrisroos:
This is similar to the change in [this commit](https://github.com/freerange/mocha/commit/8f58eddf0ff658ad255cf60cedab3c767bbb15c7).

[This commit](https://github.com/freerange/mocha/commit/e87c03b068efc48267fbcd5a295514077c52b901) made `ModuleMethod#method_exists?` redundant by changing the implementation of `ClassMethod#hide_original_method` so that it uses `ClassMethod#method_visibility` instead.

This introduced a subtle change in behaviour due to the difference in `ModuleMethod#method_exists?` and `ClassMethod#method_visibility`. The former would return false if the method being stubbed was on `Kernel` or `Object` while the latter will return true in those cases. [This gist](https://gist.github.com/chrisroos/12a1b032b95664448c9e987132f33988)
illustrates the difference.

Practically, this meant that _some_ stubbed module methods (those defined on Kernel and Object) wouldn't have had the correct visibility set. The tests added in this commit ensure we don't introduce
regressions in future. The tests for the visibility of the protected and private methods failed prior to [this commit](https://github.com/freerange/mocha/commit/e87c03b068efc48267fbcd5a295514077c52b901). I confirmed this by checking our the commit before that (`git co e87c03b068efc48267fbcd5a295514077c52b901~`) and running these new tests.

### DIFF
--- a/lib/mocha/module_method.rb
+++ b/lib/mocha/module_method.rb
@@ -3,14 +3,6 @@ require 'mocha/class_method'
 module Mocha
 
   class ModuleMethod < ClassMethod
-
-    def method_exists?(method)
-      return true if stubbee.public_methods(false).include?(method)
-      return true if stubbee.protected_methods(false).include?(method)
-      return true if stubbee.private_methods(false).include?(method)
-      return false
-    end
-
   end
 
 end
--- a/test/acceptance/stub_instance_method_defined_on_kernel_module_test.rb
+++ b/test/acceptance/stub_instance_method_defined_on_kernel_module_test.rb
@@ -72,4 +72,67 @@ class StubInstanceMethodDefinedOnKernelModuleTest < Mocha::TestCase
   ensure
     Kernel.module_eval { remove_method :my_instance_method }
   end
+
+  def test_should_stub_public_module_method_and_leave_it_unchanged_after_test
+    Kernel.module_eval do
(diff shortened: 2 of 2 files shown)


chrisroos:
I'm fairly confident this can just be merged but I don't have time to look at it right now so I'm opening this PR as a reminder that I need to do something with it.

floehopper:
@chrisroos:

This looks good to me. I'm going to remove the `wem.md` and `wem.rb` files from the commit, rebase this against `master` and force-push in preparation for merging. Let me know if you think we need to check anything else before merging or releasing this change.

chrisroos:
Travis reported a failing test. I suspect it was an intermittent failure so I've requested the build to be retried.

chrisroos:
> This looks good to me. I'm going to remove the wem.md and wem.rb files from the commit, rebase this against master and force-push in preparation for merging. Let me know if you think we need to check anything else before merging or releasing this change.

@floehopper. I've had a look at this change and am fairly confident it's safe to merge if/when the Travis build passes. I'll rebase it again and get it merged.

floehopper:
@chrisroos Sounds good. Please go ahead and merge it.

PR #297, opened 2017-03-21
Add Ruby 2.4 and ruby-head in .travis.yml

junaruga:
I want to add ruby-2.4 and ruby-head in `.travis.yml`.

## Ruby 2.4

I added "2.4.0", not "2.4".
Because last month I tried "2.4" in `.travis.yml`, it was not available.
I can not find official document for that.

## ruby-head

I want to add `ruby-head` as allow_failures in `.travis.yml`.
I think this is useful.
Because we can prepare before next version Ruby 2.5 release.
Though it might not be useful until the actual preview release.
We can support the next version as faster.

We can see this kind of logic in `rails`, `rspec` and `cucumber` and etc.
I think that is the reason why `rails` and `rspec` can support new version Ruby as faster.

https://github.com/rails/rails/blob/master/.travis.yml
https://github.com/rspec/rspec-core/blob/master/.travis.yml
https://github.com/cucumber/cucumber-ruby/blob/master/.travis.yml
https://github.com/sparklemotion/nokogiri/blob/master/.travis.yml

Is it possible to add it?

Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ rvm:
   - 2.2.6
   - 2.3.3
   - 2.4.0
+  - ruby-head
   - jruby
 gemfile:
   - Gemfile
@@ -18,6 +19,9 @@ env:
 matrix:
   include:
     - rvm: jruby
+      gemfile: gemfiles/Gemfile.minitest.latest
+      env: MOCHA_OPTIONS=debug MOCHA_RUN_INTEGRATION_TESTS=minitest
+    - rvm: ruby-head
       gemfile: gemfiles/Gemfile.minitest.latest
       env: MOCHA_OPTIONS=debug MOCHA_RUN_INTEGRATION_TESTS=minitest
     - rvm: 2.4.0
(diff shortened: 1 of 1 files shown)


floehopper:
Thanks. I'd be happy to accept this PR if you split it up into multiple commits as follows:

1. Convert all *existing* MAJOR.MINOR Ruby versions to latest MAJOR.MINOR.PATCH as suggested in travis-ci/travis-ci#7077. Please reference this issue in the commit message.
2. Additions for Ruby v2.4.0 - it needs to be added [here](https://github.com/freerange/mocha/blob/master/.travis.yml#L5), [here](https://github.com/freerange/mocha/blob/master/.travis.yml#L22-L24), and [here](https://github.com/freerange/mocha/blob/master/.travis.yml#L85-L87). Your PR missed the third of these, i.e. the one for `Test::Unit`.
3. Additions for `ruby-head`. I think this should be added in the same way, i.e. in 3 places - or at least for both `Minitest` and `Test::Unit`. Also explain why you've added `allow_failures` and `fast_finish` options in the commit note.

junaruga:
@floehopper hi, ok let me try it. I missed `test-unit` case.

floehopper:
@junaruga: Thanks. Merging.

PR #303, opened 2017-06-18
handle cases where parameter value is blank for query string matcher

weynsee:
Currently the following will raise an exception:
```
object = mock()
object.expects(:method_1).with(has_equivalent_query_string('http://example.com/foo?a='))
object.method_1('http://example.com/foo?a=')
```
This PR adds a fix so that the above assertion will pass.


floehopper:
@weynsee: Thanks. I've done some tweaking of your changes and opened #304 in place of this. Closing.

PR #305, opened 2017-06-18
Use faster container-based virtual environments for Travis CI builds

floehopper:
`sudo: required` is [the default for repositories before 2015][1].

Builds for repositories with `sudo: false` will be run in container-based virtual environments which have a much faster boot time.

[1]: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 before_install:
   - gem install bundler
 bundler_args: "--verbose"


floehopper:
Comparison of boot times:

* [Build 478.1](https://travis-ci.org/freerange/mocha/jobs/243728889#L5) (before): 21.746450374s
* [Build 483.1](https://travis-ci.org/freerange/mocha/jobs/244192962#L5) (after): 1.155293469s

PR #316, opened 2017-09-24
Ruby 2.5 Warnings

amatsuda:
Ruby 2.5 warns stricter about unused variables (since r59585).
So here's a fix.

### DIFF
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -33,11 +33,11 @@ module Mocha
 
       def method_exists?(method, include_public_methods = true)
         if include_public_methods
-          return true if @stubba_object.public_instance_methods(include_superclass_methods = true).include?(method)
+          return true if @stubba_object.public_instance_methods(_include_superclass_methods = true).include?(method)
           return true if @stubba_object.allocate.respond_to?(method.to_sym)
         end
-        return true if @stubba_object.protected_instance_methods(include_superclass_methods = true).include?(method)
-        return true if @stubba_object.private_instance_methods(include_superclass_methods = true).include?(method)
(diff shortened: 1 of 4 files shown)


floehopper:
@amatsuda: Thanks for your pull request. Has Ruby 2.5 been released yet? And do you know whether Travis CI supports it? I'd prefer to see the warnings in a Travis CI build so that I can be confident that these changes fix them.

amatsuda:
@floehopper 
> Has Ruby 2.5 been released yet?

No, Ruby's stable release is scheduled around Christmas this year again.

> And do you know whether Travis CI supports it?

Yes, you can compile ruby trunk and test against it on Travis via `ruby-head` (see https://github.com/kaminari/kaminari/blob/508009c4/.travis.yml#L6 for an actual example).
But please don't expect ruby trunk to always be stable. So don't forget to add it to `allow_failures` [like this](https://github.com/kaminari/kaminari/blob/508009c4/.travis.yml#L51-L52) when you're actually adding it to .travis.yml here.

floehopper:
@amatsuda:

I'm sorry for taking so long to respond. I've just added Ruby v2.5.0 to the Travis CI build matrix and the warnings you mention do not seem to appear in either of the relevant build jobs:

* https://travis-ci.org/freerange/mocha/jobs/325762315
* https://travis-ci.org/freerange/mocha/jobs/325762371

I can only assume some more changes were made to Ruby before the v2.5.0 release which mean that these cases no longer generate warnings.

Anyway, on the basis that it is no longer needed, I'm going to close this PR. Thanks again for taking the time to submit the PR. Please let me know if you disagree with this decision.

PR #320, opened 2018-01-28
move minitest to the correct namespace

grosser:
had to open the gem to figure out why `require 'mocha/minitest'` did not work :(
... just a quickfix, proper move should happen in the next major release ...

@floehopper

### DIFF
--- a/lib/mocha/mini_test.rb
+++ b/lib/mocha/mini_test.rb
@@ -1,3 +1,2 @@
-require "mocha/integration/mini_test"
-
-Mocha::Integration::MiniTest.activate
+warn "Require mocha/minitest instead of mocha/mini_test"
+require 'mocha/mini_test'
--- a/lib/mocha/minitest.rb
+++ b/lib/mocha/minitest.rb
@@ -1,0 +1,3 @@
+require "mocha/integration/mini_test"
+
+Mocha::Integration::MiniTest.activate


floehopper:
@grosser:

Thanks for the PR. I can see why this is surprising, although, in my defence, I think the problem originally lay with minitest, because the namespace [was originally `MiniTest`][1]. Also the correct `require` statement is mentioned several times in the Mocha README, e.g. [here][2], [here][3] & [here][4], so no need to open the gem source.

I think I might fix the problem by adding the `mocha/mini_test.rb` file, but leaving `mocha/minitest.rb` in place, but with a deprecation warning. What do you think?

[1]: https://ruby-doc.org/stdlib-2.1.0/libdoc/minitest/rdoc/MiniTest.html
[2]: https://github.com/freerange/mocha/blob/0426e5ec119bc6a6c0ab256c2b3a8273d798a6c0/README.md#minitest
[3]: https://github.com/freerange/mocha/blob/0426e5ec119bc6a6c0ab256c2b3a8273d798a6c0/README.md#minitest-1
[4]: https://github.com/freerange/mocha/blob/0426e5ec119bc6a6c0ab256c2b3a8273d798a6c0/README.md#minitest-2

grosser:
you mean adding a deprecation to mini_test.rb ? (there already is a
mini_test.rb)

floehopper:
Yes - sorry I got that the wrong way round. I should have said:

> I think I might fix the problem by adding a `mocha/minitest.rb` file, but leaving `mocha/mini_test.rb` in place, but with a deprecation warning and requiring the new `mocha/minitest.rb` file to avoid duplication. What do you think?

grosser:
ok, updated

floehopper:
Addressed in #322. Closing.

floehopper:
@grosser I've just released [v1.4.0](https://github.com/freerange/mocha/compare/v1.3.0...v1.4.0) which includes [the fix for this](https://github.com/freerange/mocha/issues/324). Let me know if you run into any problems with it.

grosser:
works!

PR #323, opened 2018-02-12
Define respond_to_missing? on Mocha::Mock

### LINKED ISSUE #321: Warning when delegating to mock in Ruby 2.4
Ruby 2.4 warns when delegating to a private method with `Forwardable`: https://bugs.ruby-lang.org/projects/ruby-trunk/repository/revisions/56210.

When delegating to a Mocha mock object with `Forwardable`, a "forwarding to private method" warning is shown.

For example, given:

```ruby
require "forwardable"

class Foo
  extend Forwardable

  attr_reader :baz

  def_delegator :baz, :bar

  def initialize(baz)
    @baz = baz
  end
end
```

```ruby
require "minitest/autorun"
require "mocha/mini_test"

require_relative "foo"

  def test_delegates_bar_to_baz
    baz = mock
    baz.expects(:bar)

    foo = Foo.new(baz)
    foo.bar
  end
end
```

The test will output:

```
foo_test.rb:12:in `test_delegates_bar_to_baz': Foo#bar at /Users/thomcarter/code/mocha/foo.rb:8 forwarding to private method Mocha::Mock#bar
```

Ideally these warnings would not be shown, because in most cases the expectation is set on a public method.

tjvc:
Fixes #321.

As of Ruby 2.4, `Forwardable` warns when delegating to a private method.
Warnings are shown when delegating methods to a `Mocha::Mock`.

To determine whether to show a warning, Forwardable tests the delegated
method with the `defined?` operator. Mocha uses `method_missing` to respond
to mocked methods. `defined?` uses `respond_to_missing?` to determine
whether or not a missing method is defined. Defining `respond_to_missing?`
on `Mocha::Mock`, in place of `respond_to?`, prevents warnings from being
shown.

To preserve compatibility with Ruby 1.8.7, we continue to conditionally
define `respond_to?`. Prior to Ruby 1.9.3, `defined?` returns `nil` for method
calls handled via `method_missing`, so we also only test that mocked
methods are `defined?` in later Ruby versions.

### DIFF
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -9,6 +9,7 @@ require 'mocha/unexpected_invocation'
 require 'mocha/argument_iterator'
 require 'mocha/expectation_error_factory'
 require 'mocha/deprecation'
+require 'mocha/ruby_version'
 
 module Mocha
 
@@ -314,7 +315,7 @@ module Mocha
     end
 
     # @private
-    def respond_to?(symbol, include_private = false)
+    def respond_to_missing?(symbol, include_private = false)
       if @responder then
         if @responder.method(:respond_to?).arity > 1
           @responder.respond_to?(symbol, include_private)
@@ -323,6 +324,13 @@ module Mocha
         end
       else
         @everything_stubbed || all_expectations.matches_method?(symbol)
+      end
+    end
+
+    # @private
+    if PRE_RUBY_V19
+      def respond_to?(symbol, include_private = false)
+        respond_to_missing?(symbol, include_private)
       end
     end
 
--- a/test/unit/mock_test.rb
+++ b/test/unit/mock_test.rb
@@ -334,6 +334,14 @@ class MockTest < Mocha::TestCase
     assert_match(/unexpected invocation/, e.message)
   end
 
+  unless PRE_RUBY_V19
+    def test_expectation_is_defined_on_mock
+      mock = build_mock
+      mock.expects(:method1)
+      assert defined? mock.method1
+    end
+  end
+
   private
 
   def build_mock


floehopper:
@tjvc Thanks for digging into this a bit more. There are already places in the code which do something different for older Ruby versions. Can I suggest you add `require 'mocha/ruby_version'` to `lib/mocha/mock.rb` and use `if PRE_RUBY_V19` to define `respond_to?` with an `else` which defines `respond_to_missing?`. Ideally I'd really like to see some kind of test for this - perhaps a test could check the same thing that `Forwardable` is checking to decide whether or not to issue a warning...?

tjvc:
Thanks for the pointer @floehopper. I've updated the code to conditionally define `respond_to?` for Ruby versions < 1.9. I've also added a test that expectations are `defined?` on mocks, which is the check that `Forwardable` uses. Because `defined?` doesn't seem to like method calls handled via `method_missing` in version 1.8.7, we only run this test for later versions. That seemed reasonable, because the behaviour is only really required for versions >= 2.4, and there are similar examples elsewhere in the code. Let me know what you think!

floehopper:
@tjvc Thanks for updating the PR. I'll try to take a look as soon as I have time.

tjvc:
Hi @floehopper, was just wondering if you'd had a chance to take a look at this yet?

floehopper:
@tjvc Sorry for the delay - we've been very busy with client work. This looks great to me! I'll merge it now.

floehopper:
I'll try to get a release out in the next few days including this change.

tjvc:
No worries, @floehopper. Thanks for merging! 👍

floehopper:
I've just released [v1.4.0](https://github.com/freerange/mocha/compare/v1.3.0...v1.4.0) which includes this fix.

PR #319, opened 2018-01-06
[WIP] Fail fast if mocking before setup

floehopper:
This is work that came out of #292.

It needs more work before it can be merged, but I thought it was worth making it into a PR to highlight the work that has already been done.

PR #327, opened 2018-03-30
Prevent use of Mocha outside the context of a test/example

### LINKED ISSUE #292: Prevent calls to Mocha::API methods until the before/setup block for a single test/example
I'm working on a relatively old Rails codebase which uses Mocha with RSpec.

There is some instability in the test suite which is caused by stubbing happening in `before(:all)` blocks:

https://gist.github.com/andyw8/7aa312287dd5c60412475a147900e121

If I attempt the same thing with rspec-mocks, it refuses to run:

`The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported.`

https://relishapp.com/rspec/rspec-mocks/docs/basics/scope#cannot-create-doubles-in-a-`before(:context)`-hook

Is it possible to have a similar level of protection when using mocha? I'm not sure if this code would belong in Mocha, in RSpec, or perhaps a separate gem.

### LINKED ISSUE #293: Outer tests in acceptance tests should not interact with Mocha
Currently both the outer and inner tests interact with Mocha (and thus the Mockery). However, only the inner tests need to interact with Mocha. By inner tests, I mean the ones created using `TestRunner#run_as_test` or `#run_as_tests`. By outer tests, I mean the acceptance tests which call these `TestRunner` methods.

This came up when I was working on #292, because the double-teardown became problematic when making those changes.

floehopper:
This change means that any attempt to use Mocha outside the context of a test or example will result in a `Mocha::NotInitializedError` being raised with the following message: "Mocha methods cannot be used outside the context of a test".

Fixes #292
Fixes #293 (indirectly)
Supersedes #319

### DIFF
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -55,9 +55,7 @@ module Mocha
 
       def teardown
         instance.teardown
-      end
-
-      def reset_instance
+      ensure
         @instances.pop
         @instances = nil if instances.empty?
       end
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -7,16 +7,40 @@ require 'mocha/state_machine'
 require 'mocha/logger'
 require 'mocha/configuration'
 require 'mocha/stubbing_error'
+require 'mocha/not_initialized_error'
 require 'mocha/expectation_error_factory'
 
 module Mocha
 
   class Mockery
 
+    class Null < self
+
+      def add_mock(*)
+        raise_not_initialized_error
+      end
+
+      def add_state_machine(*)
+        raise_not_initialized_error
+      end
+
+      def stubba
+        Central::Null.new(&method(:raise_not_initialized_error))
+      end
+
+      private
+
+      def raise_not_initialized_error
+        message = 'Mocha methods cannot be used outside the context of a test'
+        raise NotInitializedError.new(message, caller)
+      end
+
+    end
+
     class << self
 
       def instance
-        instances.last
+        instances.last || Null.new
       end
 
       def setup
@@ -35,6 +59,7 @@ module Mocha
 
       def reset_instance
         @instances.pop
+        @instances = nil if instances.empty?
       end
 
       private
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -16,7 +16,13 @@ module Mocha
     class << self
 
       def instance
-        @instance ||= new
+        instances.last
+      end
+
+      def setup
+        mockery = new
+        mockery.logger = instance.logger unless instances.empty?
+        @instances.push(mockery)
       end
 
       def verify(*args)
@@ -28,7 +34,13 @@ module Mocha
       end
 
       def reset_instance
-        @instance = nil
+        @instances.pop
+      end
+
+      private
+
+      def instances
+        @instances ||= []
       end
 
     end
--- a/lib/mocha/central.rb
+++ b/lib/mocha/central.rb
@@ -2,6 +2,23 @@ module Mocha
 
   class Central
 
+    class Null < self
+
+      def initialize(&block)
+        super
+        @raise_not_initialized_error = block
+      end
+
+      def stub(*)
+        @raise_not_initialized_error.call
+      end
+
+      def unstub(*)
+        @raise_not_initialized_error.call
+      end
+
+    end
+
     attr_accessor :stubba_methods
 
     def initialize
--- a/lib/mocha/error_with_filtered_backtrace.rb
+++ b/lib/mocha/error_with_filtered_backtrace.rb
@@ -1,0 +1,17 @@
+require 'mocha/backtrace_filter'
+
+module Mocha
+
+  # @private
+  class ErrorWithFilteredBacktrace < StandardError
+
+    # @private
+    def initialize(message = nil, backtrace = [])
+      super(message)
+      filter = BacktraceFilter.new
+      set_backtrace(filter.filtered(backtrace))
+    end
+
+  end
+
+end
--- a/lib/mocha/hooks.rb
+++ b/lib/mocha/hooks.rb
@@ -21,6 +21,7 @@ module Mocha
     #
     # This method should be called before each individual test starts (including before any "setup" code).
     def mocha_setup
+      Mockery.setup
     end
 
     # Verifies that all mock expectations have been met (only for use by authors of test libraries).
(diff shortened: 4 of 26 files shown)


lib/mocha/mockery.rb
```diff
@@ -7,16 +7,47 @@
...
+        raise NotInitializedError.new(message, caller)
+      end
+
+    end
+
     class << self
 
       def instance
-        @instance ||= new
+        @instances.last || Null.new
```
lzap: This throws `undefined method `last' for nil:NilClass` when you attempt to use mocha uninitialized (e.g. example snippets from documentation).

lzap:
Works great here, I vote for some easy opt-out, I can imagine this can break tests for some. An environmental variable should be good enough for those who don't want to fix their tests, can't or want to postpone this. Kudos, great job!

lib/mocha/mockery.rb
floehopper: @lzap Gah! Looks like I went wrong with some of my interactive rebasing. Thanks for testing - I'll try to sort it out as soon as I can.

floehopper: @lzap I'm in the process of fixing this and just to warn you I plan to force-push the branch shortly.

floehopper:
@lzap:

> Works great here

Thanks for testing it. 👍 

> I vote for some easy opt-out, I can imagine this can break tests for some. An environmental variable should be good enough for those who don't want to fix their tests, can't or want to postpone this.

I'm hesitant to offer an opt-out, because Mocha was never intended to be used outside the context of a test. I'm probably going to release it like this and see whether anyone runs into problems.

In any case, a better solution for that problem is probably to call `Mocha::Hooks#mocha_setup` from an appropriate place and at least consider also calling `Mocha::Hooks#mocha_verify` and `Mocha::Hooks#mocha_teardown` which is what the various test frameworks use. See [this documentation][1] for more info.

[1]: http://gofreerange.com/mocha/docs/Mocha/Hooks.html

floehopper:
I just ran the tests in the [GOV.UK Whitehall project](https://github.com/alphagov/whitehall) with these changes and saw no ill effects, so I'm happy to merge.

floehopper:
@lazyatom has just [updated Kintama][1] to work with this change and all appears to be well.

[1]: https://github.com/lazyatom/kintama/commit/1db062a69a760624d217c7faa6a3ef47889f2492

floehopper:
I've released this change in v1.5.0.

Evan-M:
>  I'm probably going to release it like this and see whether anyone runs into problems.

:raising_hand_man: I'm running into problems... A gem I've been contributing to now has failing tests on Travis CI for all branches. One of the tests stubs out method calls in `before` blocks of a handful of controller tests/examples; while ultimately the test in question _should_ probably be refactored, I would not have expected an updated dependency to break a bunch of tests either.  (To be fair, the `Gemfile` does not specify a version of Mocha, and Travis CI seems to have updated to `1.5.0` automagically, which I also didn't expect)

(A release with a deprecation warning instead of raising an exception might have been nice, in this case).

floehopper:
@Evan-M Sorry to hear that. Can you point me at one of the tests which is failing? What test framework are you using? There shouldn't be a problem stubbing methods in a `before` block, so that's a bit confusing.

floehopper:
@Evan-M I've left [a comment on the issue you raised on `lynndylanhurley/devise_token_auth`][1]. I hope it helps. Let me know if you have any questions.

[1]: https://github.com/lynndylanhurley/devise_token_auth/issues/1137#issuecomment-379409853

Evan-M:
Yeah, no problem... I should have linked to the failures in my original comment.

**A little more pretext:**
The first encounter with the problem was in the build status for Travis after merging lynndylanhurley/devise_token_auth#1134. Everything passed on the PR, and then started failing midway through the CI build jobs for `master` __**after**__ merging the PR.  The [job logs](https://travis-ci.org/lynndylanhurley/devise_token_auth/jobs/363204323#L765) for the build report a failing test suite, with the error:
~~~sh
Mocha::NotInitializedError: Mocha methods cannot be used outside the context of a test
~~~

When I tried to re-run the build, all the build failed; I suspect I merged lynndylanhurley/devise_token_auth#1134 right about the same time that version 1.5.0 was released.

The [same build failure / test failure](https://travis-ci.org/lynndylanhurley/devise_token_auth/jobs/363226837#L775) appears to happen for [another (unmerged) PR](https://github.com/lynndylanhurley/devise_token_auth/pull/1004) as well.

---------

>  What test framework are you using?

The test suite is based on Minitest, but there is a mixture of the Minitest Spec DSL and the Minitest Test DSL.  (It was that way when I became involved in the project 🤷‍♂️).

> There shouldn't be a problem stubbing methods in a before block, so that's a bit confusing.

Yeah, I was wondering about stubbing methods in a `before` block (as opposed to a `before(:all)` block, which doesn't even exist in Minitest!).

> Can you point me at one of the tests which is failing?

Anyway, the repo is [lynndylanhurley/devise_token_auth]. The failing tests are all in the test file  [`/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb`]:

https://github.com/lynndylanhurley/devise_token_auth/blob/7df89606ef127961c1957aaccf714ff322ebf431/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb#L72-L75

https://github.com/lynndylanhurley/devise_token_auth/blob/7df89606ef127961c1957aaccf714ff322ebf431/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb#L147-L166

https://github.com/lynndylanhurley/devise_token_auth/blob/7df89606ef127961c1957aaccf714ff322ebf431/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb#L168-L182

Also the failing tests in question here are actually part of an Integration Test, so it is probably a bad idea to stub any method calls!

[lynndylanhurley/devise_token_auth]: https://github.com/lynndylanhurley/devise_token_auth
[`/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb`]: https://github.com/lynndylanhurley/devise_token_auth/blob/7df8960/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb

floehopper:
@Evan-M The upgrade to Mocha v1.5.0 definitely revealed a problem, but I think there was already an underlying problem in that test. Did you see my comment: https://github.com/lynndylanhurley/devise_token_auth/issues/1137#issuecomment-379409853 ? If you fix that, there's shouldn't be a problem stubbing/expecting methods in a `before` block and using Mocha v1.5.0.

Issue #298, opened 2017-03-27
Expectation ignored when test cases are run in a certain order


skliew:
Found this issue using version `1.2.1`. In the code example as follows, `test_case_2` passes even when there is no call to any method named `test1`, resulting in a false positive:

```ruby
require 'test-unit'
require 'mocha/test_unit'

class SUT
  def test1
  end
end

class MyTest < Test::Unit::TestCase
  self.test_order = :defined
  def test_case_1
    SUT.any_instance.expects(:test1).once
    SUT.new.test1
  end

  def test_case_2
    Object.any_instance.expects(:test1).once
  end
end

```

Do note that when the line `SUT.new.test1` is commented out, both test cases fail. And when `test_case_2` is run before `test_case_1`, only `test_case_2` fails which should be the correct behavior.

floehopper:
@skliew: Thanks for reporting this issue. Setting an expectation on any instance of `Object` seems like a bit of a pathological case, but at first glance I agree the behaviour doesn't sound correct. We'll try to look into it as soon as we have some spare time. Out of interest, do you have some real tests which need to do this?

skliew:
I actually found this problem in a real code base when a test case occasionally fails. I have to admit that the test case was itself buggy, as most of the time false positives are reported.

And thanks for the great library :)

floehopper:
@skliew Hmm. Sorry for taking so long to get back to you. I just did a bit of digging and it appears this might be a more fundamental problem with the Test::Unit integration. I'm continuing to investigate.

floehopper:
I take back my last comment about it being a more fundamental problem. I made the mistake of looking at this before I'd had any coffee! I'm continuing to investigate.

floehopper:
@skliew: I've opened https://github.com/freerange/mocha/pull/331 as an attempt to fix this problem. It turns out that it is indeed a bit more general than I first thought, so many thanks for reporting it!

floehopper:
@skliew The fix in #331 was just released in v1.6.0.

PR #331, opened 2018-04-20
Avoid mocks for partial mocking leaking into subsequent tests

floehopper:
Previously `ObjectMethods#mocha` & `ClassMethods::AnyInstance#mocha` were building mock objects for ancestor classes when `ObjectReceiver#mocks` or `AnyInstanceReceiver#mocks` was called from `Mock#all_expectations` as part of `Mock#method_missing`.

Since these mock objects were not associated with a corresponding `ClassMethod` or `AnyInstanceMethod` registered with `Central` via `Mockery#mock_impersonating` or `Mockery#mock_impersonating_any_instance_of`, they were not being reset as part of `Central#unstub_all` via `Mockery#teardown` and this state was leaking into subsequent tests.

You can see an example of this in #298. I've captured a more general case of this in the new acceptance test method in `StubAnyInstanceMethodDefinedOnSuperclassTest` which was failing before the fix in this commit.

I've also added a similar acceptance test method to `StubClassMethodDefinedOnSuperclassTest` which was also failing before the fix in this commit.

I'm not particularly happy with the fix, but I think it'll do for now.

### DIFF
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -19,8 +19,12 @@ module Mocha
         @stubba_object = klass
       end
 
-      def mocha
-        @mocha ||= Mocha::Mockery.instance.mock_impersonating_any_instance_of(@stubba_object)
+      def mocha(instantiate = true)
+        if instantiate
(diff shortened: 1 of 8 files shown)


floehopper:
Released in v1.6.0.

PR #333, opened 2018-06-05
Use Kernel#warn for deprecations

etiennebarrie:
That makes it possible to override `Warning.warn` starting with Ruby 2.5.0

See more at https://bugs.ruby-lang.org/issues/12299

### DIFF
--- a/lib/mocha/deprecation.rb
+++ b/lib/mocha/deprecation.rb
@@ -13,7 +13,7 @@ module Mocha
         unless mode == :disabled
           filter = BacktraceFilter.new
           location = filter.filtered(caller)[0]
-          $stderr.puts "Mocha deprecation warning at #{location}: #{message}"
+          warn "Mocha deprecation warning at #{location}: #{message}"
         end
       end
 


etiennebarrie:
This only applies to deprecations, if you're interested I can look at implementing this for the mocking warnings as well.
There were no tests for `Mocha::Deprecation` so I didn't add one specifically but I think it could be done even if it changes a global, because the tests don't run in parallel.

floehopper:
I've only had a chance to read this very quickly and I'm not sure I understand. Can you give some example code demonstrating the problem you are trying to solve?

etiennebarrie:
Using `$stderr.puts` does not respect the Ruby conventions which allow to set `-W0` to silence warnings:
```sh
$ ruby -W0 -e 'warn "warning: not shown"'
# no output
$ ruby -e 'warn "warning: shown"'
warning: shown
```

https://idiosyncratic-ruby.com/3-ruby-can-you-speak-louder.html#command-line-options-for-debug-modes 
https://mislav.net/2011/06/ruby-verbose-mode/

Using `Kernel#warn` gets us this feature for free, so we could remove `Mocha::Deprecation.mode` and change `$VERBOSE` directly in `DeprecationDisabler.disable_deprecations`.

---

Additionally, since Ruby 2.5.0, `Kernel#warn` calls `Warning.warn` in verbose mode, which lets you override it as described in feature 12299: https://bugs.ruby-lang.org/issues/12299

```sh
$ ruby -W0 -e 'def Warning.warn(warning) super("warning: #{warning}") end' -e 'warn "not shown"'
# no output
$ ruby -e 'def Warning.warn(warning) super("warning: #{warning}") end' -e 'warn "shown"'
warning: shown
```

I don't know if it's a good idea really, but my use case is to raise for warnings in CI/dev in a Rails app:

```ruby
unless Rails.env.production?
  def Warning.warn(warning)
    raise warning
  end
end
```

etiennebarrie:
Fixed in 196970ac6c972bae612fb17f8afdee443d700c01. Thanks!

floehopper:
@etiennebarrie Oh, cool! Sorry for not responding sooner and thanks for noticing the (accidental!) fix. I'll try to get a release including the fix shipped in the not too distant future.

etiennebarrie:
No problem. Thank you for your work!

PR #335, opened 2018-06-22
Change gem version badge to svg and add SemVer stability badge

greysteil:
First of all, thanks for Mocha!

I noticed the gem version badge on your readme was using to old png format, so I've updated it to be an svg. I've also added a badge indicating how bug-free / non-breaking new releases are, based on all of the updates of Mocha that dependabot has done (I built it). Hopefully that's helpful, too, but feel free to just take the svg commit. :octocat: 

[Gem Version](http://badge.fury.io/rb/mocha) [SemVer](https://dependabot.com/compatibility-score.html?dependency-name=mocha&package-manager=bundler&version-scheme=semver)


floehopper:
Released in v1.6.0.

PR #336, opened 2018-07-20
Cd/fcv 4941b/fix error message in safe stubs

stripedpumpkin:


floehopper:
@stripedpumpkin, @gleseur: I saw this pull request (which you've since closed) and glanced through the changes. I wonder whether you might be able to use the built-in configuration option `:stubbing_non_existent_method` with `#stubs` and `#expects` instead of  the new`#safe_stubs` and `#safe_expects` methods which you've added. This way you wouldn't need to maintain a fork of Mocha. See [this documentation](http://gofreerange.com/mocha/docs/Mocha/Configuration.html) for further details.

stripedpumpkin:
@floehopper Interesting, we look into it. What happens if you used this configuration and use ``#stubs`` on a ``mock`` object?

floehopper:
@stripedpumpkin: I believe the use of `#stubs` on a mock object should be unaffected by this configuration option.

stripedpumpkin:
@floehopper Another thing we have been considering was to have a shorter way of writing
```ruby
def fake_sheep
  @fake_sheep ||= mock('sheep').tap do |sheep|
    sheep.responds_like_instance_of(Sheep)
  end
end
```
as something like (naming to be discussed)
```ruby
def fake_sheep
  @fake_sheep ||= mocking(Sheep, 'sheep')
end
```
The idea is again to enforce the mock to indicate what it is supposed to be mocking. That would add an additional wrapper around a mock, but that would be a step towards making it easy to enforce such safer mocks. Ultimately, a good thing would be to have a config that raise if you ``#expects`` or ``#stubs`` on a mock that has not declared what it is, it is supposed to be mocking. The first part we could do within our code base I suppose, but the second would probably be best as a Mocha config. Maybe there is already something like that in mocha?

floehopper:
@stripedpumpkin There is already a shorter way of writing the `#fake_sheep` implementation:

```ruby
def fake_sheep
  @fake_sheep ||= mock('sheep').responds_like_instance_of(Sheep)
end
```

The methods like `#responds_like_instance_of` return the mock so that you can chain the method calls. I've added [a new issue](https://github.com/freerange/mocha/issues/337) to improve the documentation to make this possibility clearer. While I realise this isn't as terse as your suggested new `#mocking` method, hopefully it's good enough.

> Ultimately, a good thing would be to have a config that raises if you #expects or #stubs on a mock that has not declared what it is, it is supposed to be mocking.

That's an interesting idea. I've captured it in [another new issue](https://github.com/freerange/mocha/issues/338).

If you have any further thoughts, please could you comment on the new issues (rather than on this one) or if appropriate open new ones...? Thanks.

PR #343, opened 2018-10-06
Fix object inspection calling instance methods on object

### LINKED ISSUE #342: Object.mocha_inspect not compatible with classes that define a 'format' method
It looks like setting an expectation on a class that defines a `format` method results in an `ArgumentError`. 

Take the following test case:

```
class ObjectTest < ActiveSupport::TestCase

  class DummyClass
    attr_reader :format
  end

  test 'dummy class expectations' do
    dummy_class = DummyClass.new
    dummy_class.expects(:foo)
  end

end
```

Running it:

```
ObjectTest
  test_dummy_class_expectations                                  ERROR (0.13s)
ArgumentError:         ArgumentError: wrong number of arguments (2 for 0)

```

This is because `mocha_inspect` calls the `format` method:

```
def mocha_inspect
  address = __id__ * 2
  address += 0x100000000 if address < 0
  inspect =~ /#</ ? "#<#{self.class}:0x#{format('%x', address)}>" : inspect
end
```

Would you consider changing that to call `Kernel.format` directly? This way it won't interfere with any instance methods on the object being inspected. If so, let me know and I'll be happy to put together a PR.

hoffmanilya:
This fixes #342 by calling `Kernel.format` directly when inspecting an object.

I also noticed the same formatting present in `expectation.rb` (line 586) and 'names.rb' (line 40). Since these don't look to be prepended to any objects I'm assuming they're safe as they are but please let me know if that isn't the case.

### DIFF
--- a/lib/mocha/inspect.rb
+++ b/lib/mocha/inspect.rb
@@ -5,7 +5,7 @@ module Mocha
     def mocha_inspect
       address = __id__ * 2
       address += 0x100000000 if address < 0
-      inspect =~ /#</ ? "#<#{self.class}:0x#{format('%x', address)}>" : inspect
+      inspect =~ /#</ ? "#<#{self.class}:0x#{Kernel.format('%x', address)}>" : inspect
     end
   end
 
--- a/test/unit/object_inspect_test.rb
+++ b/test/unit/object_inspect_test.rb
@@ -49,7 +49,7 @@ class ObjectInspectTest < Mocha::TestCase
   def test_should_not_call_object_instance_methods
     object = Object.new
     class << object
-      def format(*args)
+      def format(*)
         'internal_format'
       end
     end
--- a/test/unit/object_inspect_test.rb
+++ b/test/unit/object_inspect_test.rb
@@ -45,4 +45,14 @@ class ObjectInspectTest < Mocha::TestCase
 
     assert_equal [:__id__], calls.uniq
   end
+
+  def test_should_not_call_object_instance_methods
+    object = Object.new
+    class << object
+      def format(*args)
+        'internal_format'
+      end
+    end
+    assert_no_match(/internal_format/, object.mocha_inspect)
+  end
 end


hoffmanilya:
@floehopper - I wasn't able to run the build matrix locally.

I tried installing all the ruby version listed in `.travis.yml` but `rbenv` fails when trying to install `ruby-head` or `jruby-head`. E.g.

```
rbenv install ruby-head
ruby-build: definition not found: ruby-head
```

Temporarily removing those entries from the travis configuration fails with a different error:

```
bin/build-matrix
["2.5.1", "gemfiles/Gemfile.minitest.latest", {"MOCHA_OPTIONS"=>"debug", "MOCHA_RUN_INTEGRATION_TESTS"=>"minitest"}]
bin/build-matrix:8:in `block in execute': undefined method `success?' for nil:NilClass (NoMethodError)
```

I believe this is because i have `RVM` installed and it's preventing `rbenv` from setting the correct version. Any ideas on how to temporarily disable `RVM` and use `rbenv` instead?

Thanks!

floehopper:
@hoffmanilya:

Thanks for opening this PR and for being so thorough that you want to run the build matrix locally. I haven't used the build matrix script in a while and so I'm not surprised it has stopped working. I hope this hasn't caused you too much work.

I've tended to use Travis CI itself to run the matrix of builds in the `.travis.yml` file. In fact your PR already has [a build on Travis CI](https://travis-ci.org/freerange/mocha/builds/438058291). Although the build has failed, I suspect the failures were all to do with a Rubocop upgrade which I've [just fixed][1] on `master`. If you rebase your branch against `master` and force-push, I think the Travis CI build will pass.

I've made a note to do something about the `build-matrix` script in [a new issue][2].

[1]: https://github.com/freerange/mocha/commit/05e507f53c1e2713d8ebc4b5a0bb4f074822131d
[2]: https://github.com/freerange/mocha/issues/344

floehopper:
@hoffmanilya:

I'm sorry - it appears that my "fix" introduced another problem which is what you're seeing in your PR builds like this one: https://travis-ci.org/freerange/mocha/jobs/438069302.

I've attempted [another fix][1], but it might be worth waiting to see whether [this build][2] passes before rebasing again.

[1]: https://github.com/freerange/mocha/commit/0914c875f0da93cb9e8eb6017e8c4b42f2ab5b5a
[2]: https://travis-ci.org/freerange/mocha/builds/438071903

hoffmanilya:
@floehopper - no worries! Thanks for your help on this. I'll wait for your new build to pass and then try again.

floehopper:
I hope you don't mind, but I've squashed your two commits into one, changed the name of the new unit test, made some changes to the commit message, changed the name of the branch, and opened a new PR: #345. I'm going to close this PR in favour of the new one.

PR #347, opened 2018-10-26
Add documentation and tests for mock(:foo)

XrXr:
Becuse of the implementation of API#mock, it is possible to get a
quick stub by passing a symbol into it.

This commit adds documentation and tests for this usage.

----

I feel maybe this is undocumented until now since maybe this is not a feature we want to support. Feel free to reject this PR if that's the case.

### DIFF
--- a/lib/mocha/api.rb
+++ b/lib/mocha/api.rb
@@ -23,6 +23,7 @@ module Mocha
     #
     # @param [String] name identifies mock object in error messages.
     # @param [Hash] expected_methods_vs_return_values expected method name symbols as keys and corresponding return values as values - these expectations are setup as if {Mock#expects} were called multiple times.
+    # @param [Symbol] expected_method_name name of a method expected to be called once.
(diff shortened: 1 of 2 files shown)


floehopper:
@XrXr Sorry for the slow response. Thanks for noticing this and for opening the PR. This behaviour is completely accidental (!) which is why it isn't documented or tested. My initial reaction is that while the style is nice and terse, it's potentially easily confused with the named mock style where the argument is a String. I need to give the idea some thought. Please bear with me...

XrXr:
👋 @floehopper did you get get a chance to make a decision about this?

floehopper:
@XrXr:

Sorry for the very slow response. Having mulled it over a bit, my thinking hasn't changed since my initial reaction:

> while the style is nice and terse, it's potentially easily confused with the named mock style where the argument is a String.

So I'm afraid I'm going to reject this PR. I've opened [a new issue](https://github.com/freerange/mocha/issues/353) to deprecate or prevent this usage.

Thanks again for your suggestion and for taking the time to open a PR.

PR #354, opened 2019-03-18
Run tests in TruffleRuby in Travis CI

deepj:


floehopper:
@deepj Thanks for the PR. Are you involved in TruffleRuby development? I'm thinking of adding a section to the README about which Rubies are supported. I see that all the Mocha tests pass on TuffleRuby - can you think of any reason I shouldn't say Mocha is currently working on TruffleRuby?

deepj:
@floehopper No, not directly, but I've already reported almost 110 issues there https://github.com/oracle/truffleruby/issues/created_by/deepj

TruffleRuby supports widely Ruby. There is mainly issues with C extensions running in managed environment of Graal/Sulong VM. Then there are some "edge" use cases when the support is not full yet. I help dry-rb's gems, ROM, or Sequel to be fully supported and working in TruffleRuby. There is no reason why Mocha wouldn't work in TruffleRuby but for sure I allowed to fail since it's still in RC state.

floehopper:
@deepj Thanks for the explanation. I'm going to merge this now.

PR #357, opened 2019-06-06
Expand explanation of thread-safety concerns

techbelly:


floehopper:
Thanks, Ben. That's definitely a clearer explanation.

PR #361, opened 2019-06-09
WIP: Quick attempt at allowing tests to run in different threads

floehopper:
I ran the sample tests, which were parallelized using minitest/hell,
using the following command:

    N=10 bundle exec ruby multithreaded_test.rb

* These changes assume the setup/test/teardown sequence is run in the
same thread for any given test - it's not an attempt to allow arbitrary
multi-threading within a single test.

* Although at first glance it appears that these changes cover the most
common scenarios, I strongly suspect there are some edge cases to do
with stubbed methods potentially being defined multiple times on the
same class. This needs further investigation.

/cc @zenspider.

### DIFF
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -17,10 +17,11 @@ module Mocha
       end
 
       def mocha(instantiate = true)
+        @mochas ||= {}
         if instantiate
-          @mocha ||= Mocha::Mockery.instance.mock_impersonating_any_instance_of(@stubba_object)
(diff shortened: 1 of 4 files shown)


floehopper:
@zenspider 

So that makes it sound as if you want the code under test to be running in one or more threads that are potentially *different* to the thread the test is running in. Is that correct?

To help me understand, can you give me a simple example of a test which demonstrates the unexpected behaviour you describe?

floehopper:
I'm going to close this for now. It was an interesting experiment, but it would need more work to get it merged.

Issue #362, opened 2019-06-16
Simplify/consistent-ify ClassMethods#method_exists? & ObjectMethods#method_exists?


floehopper:
Can we re-use the code in the latest incarnation of `ClassMethod#method_visibility` or at least use a similar approach, i.e. use calls to `public|protected|private_method_defined?` vs checking for including in `public|protected|private_instance_methods` & `public|protected|private_methods`.

Issue #270, opened 2016-09-29
Multiple `method_exists?` implementations


chrisroos:
Commit e87c03b068efc48267fbcd5a295514077c52b901 changed `ClassMethod#method_exists?` to `ClassMethod#method_visibility`. PR #262 makes the same change to `AnyInstanceMethod`.

These changes have highlighted that there are a number of other definitions of `method_exists?` in the code. We should investigate whether these can be removed/replaced based on our the changes mentioned above.

PR #370, opened 2019-09-29
Simplify/consistent-ify ClassMethods#method_exists? & ObjectMethods#method_exists?

nitishr:
As suggested in #362.

Also addresses #270.

In fact, went a step further and got rid of `ClassMethod#method_visibility` as they were all essentially doing the same thing.

Similarly, the method doesn't exist anymore in `ObjectMethods`. `method_exists?` actually should anyway be available only on a `Class` instance, not on any `Object` instance.

Reduced some duplication in the action allow/prevent/warn check in `Mockery` as well, prompted by rubocop complaining about the combinatorial complexity of `on_stubbing` on the addition of just one expression to an `if` conditional.

Might be easier to review commit-by-commit than using just the diff.

### DIFF
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -86,10 +86,10 @@ module Mocha
       "#{stubbee}.#{method_name}"
     end
 
-    def method_visibility
-      original_method_owner.method_visibility(method_name)
+    def method_exists?
+      original_method_owner.method_exists?(method_name)
     end
-    alias_method :method_defined_in_stubbee_or_in_ancestor_chain?, :method_visibility
+    alias_method :method_defined_in_stubbee_or_in_ancestor_chain?, :method_exists?
 
     private
 
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -36,7 +36,7 @@ module Mocha
     end
 
     def hide_original_method
-      return unless method_defined_in_stubbee_or_in_ancestor_chain?
+      return unless method_exists?
       store_original_method_visibility
       if use_prepended_module_for_stub_method?
         use_prepended_module_for_stub_method
@@ -89,7 +89,6 @@ module Mocha
     def method_exists?
       original_method_owner.method_exists?(method_name)
     end
-    alias_method :method_defined_in_stubbee_or_in_ancestor_chain?, :method_exists?
 
     private
 
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -87,9 +87,7 @@ module Mocha
     end
 
     def method_visibility
-      (original_method_owner.public_method_defined?(method_name) && :public) ||
-        (original_method_owner.protected_method_defined?(method_name) && :protected) ||
-        (original_method_owner.private_method_defined?(method_name) && :private)
+      original_method_owner.method_visibility(method_name)
     end
     alias_method :method_defined_in_stubbee_or_in_ancestor_chain?, :method_visibility
 
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -101,7 +101,7 @@ module Mocha
     attr_reader :original_method, :original_visibility
 
     def store_original_method_visibility
-      @original_visibility = method_visibility
+      @original_visibility = original_method_owner.method_visibility(method_name)
     end
 
     def stub_method_overwrites_original_method?
--- a/lib/mocha/class_method.rb
+++ b/lib/mocha/class_method.rb
@@ -36,7 +36,7 @@ module Mocha
     end
 
     def hide_original_method
-      return unless method_exists?
+      return unless original_method_owner.method_exists?(method_name)
       store_original_method_visibility
       if use_prepended_module_for_stub_method?
         use_prepended_module_for_stub_method
@@ -84,10 +84,6 @@ module Mocha
 
     def to_s
       "#{stubbee}.#{method_name}"
-    end
-
-    def method_exists?
-      original_method_owner.method_exists?(method_name)
     end
 
     private
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -60,5 +60,6 @@ module Mocha
         (protected_method_defined?(method) && :protected) ||
         (private_method_defined?(method) && :private)
     end
+    alias_method :method_exists?, :method_visibility
   end
 end
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -33,11 +33,17 @@ module Mocha
       def method_exists?(method, include_public_methods = true)
         if include_public_methods
           return true if @stubba_object.public_instance_methods(true).include?(method)
-          return true if @stubba_object.allocate.respond_to?(method.to_sym)
+          return true if respond_to?(method)
         end
         return true if @stubba_object.protected_instance_methods(true).include?(method)
         return true if @stubba_object.private_instance_methods(true).include?(method)
         false
+      end
+
+      private
+
+      def respond_to?(method)
+        @stubba_object.allocate.respond_to?(method.to_sym)
       end
     end
 
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -35,9 +35,7 @@ module Mocha
           return true if @stubba_object.public_method_defined?(method)
           return true if respond_to?(method)
         end
-        return true if @stubba_object.protected_method_defined?(method)
-        return true if @stubba_object.private_method_defined?(method)
-        false
+        @stubba_object.protected_method_defined?(method) || @stubba_object.private_method_defined?(method)
       end
 
       private
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -56,5 +56,11 @@ module Mocha
       end
       @any_instance ||= AnyInstance.new(self)
     end
+
+    def method_visibility(method, include_public_methods = true)
+      (include_public_methods && public_method_defined?(method) && :public) ||
+        (protected_method_defined?(method) && :protected) ||
+        (private_method_defined?(method) && :private)
+    end
   end
 end
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -32,13 +32,11 @@ module Mocha
         @stubba_object.allocate.respond_to?(method.to_sym)
       end
 
-      attr_reader :stubba_object
-
-      private
-
       def singleton_class
         @stubba_object
       end
+
+      attr_reader :stubba_object
     end
 
     # @return [Mock] a mock object which will detect calls to any instance of this class.
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -32,11 +32,11 @@ module Mocha
 
       def method_exists?(method, include_public_methods = true)
         if include_public_methods
-          return true if @stubba_object.public_instance_methods(true).include?(method)
+          return true if @stubba_object.public_method_defined?(method)
           return true if respond_to?(method)
         end
-        return true if @stubba_object.protected_instance_methods(true).include?(method)
-        return true if @stubba_object.private_instance_methods(true).include?(method)
+        return true if @stubba_object.protected_method_defined?(method)
+        return true if @stubba_object.private_method_defined?(method)
         false
       end
 
--- a/lib/mocha/class_methods.rb
+++ b/lib/mocha/class_methods.rb
@@ -30,14 +30,6 @@ module Mocha
 
       attr_reader :stubba_object
 
-      def method_exists?(method, include_public_methods = true)
-        if include_public_methods
-          return true if singleton_class.public_method_defined?(method)
-          return true if respond_to?(method)
-        end
-        singleton_class.protected_method_defined?(method) || singleton_class.private_method_defined?(method)
-      end
-
       private
 
       def singleton_class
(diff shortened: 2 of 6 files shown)


floehopper:
@nitishr Thanks so much for tackling one of the outstanding issues - very much appreciated! I need to find a chunk of quiet time to review your changes, but I'm a bit busy at the moment. I'll try to get back to you as soon as I can.

nitishr:
The build now failing for ruby 1.8.7 + minitest latest.
```
    - rvm: 1.8.7-p371
      gemfile: gemfiles/Gemfile.minitest.latest
      env: MOCHA_OPTIONS=debug MOCHA_RUN_INTEGRATION_TESTS=minitest
```

This is because the latest minitest release doesn't support ruby 1.8.7. Specifically, the regex [here](https://github.com/seattlerb/minitest/blame/15ed8e4ce504c8313058a1d6fc4918299be34328/lib/minitest/assertions.rb#L139-L140) isn't recognized by ruby 1.8.7. 

> decided to lower support to include ruby 2.2.

From the above comment from the minitest [History document](https://github.com/seattlerb/minitest/blob/master/History.rdoc), it'd seem that they're dropping support for ruby < 2.2.

I suggest we update the travis config to test against at the most version `5.11.3` of minitest. The versions after that don't seem compatible.

floehopper:
@nitishr I've now fixed the minitest compatibility issue in #373. Can you rebase this branch against `master` so we can see whether all the tests pass. Thanks again for your work on this.

test/unit/any_instance_method_test.rb
```diff
@@ -6,9 +6,16 @@
 class AnyInstanceMethodTest < Mocha::TestCase
   include Mocha
 
+  def class_with_method(method, result = nil)
+    Class.new do
+      extend ClassMethods
```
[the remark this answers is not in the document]
floehopper: Can you explain why classes in this test now need to `extend Mocha::ClassMethods`? I'm concerned that this makes it less of a unit test and more of an integration test, but maybe it's OK...?

test/unit/any_instance_method_test.rb
```diff
@@ -6,9 +6,16 @@
...
+      define_method(method) { result } if method
+    end
+  end
+
```
floehopper: I'd prefer that the `class_with_method` method was initially extracted in a separate preparatory commit. I think that would make it clearer what was actually changing in this commit.

test/unit/any_instance_method_test.rb
```diff
@@ -18,14 +25,14 @@ def test_should_hide_original_method
   end
 
   def test_should_not_raise_error_hiding_method_that_isnt_defined
-    klass = Class.new
+    klass = class_with_method(nil)
```
floehopper: I think this test is now a bit confusing, because in this case `klass` is actually a class *without* a method! It might be better to have a `new_class` method which yields a block and just replaces `Class.new`.

test/unit/class_method_test.rb
```diff
@@ -4,13 +4,22 @@
...
 class ClassMethodTest < Mocha::TestCase
   include Mocha
 
+  def class_with_method(method, result = nil)
+    Class.new do
+      extend ClassMethods
+      singleton_class.extend(ClassMethods)
+      singleton_class.define_method(method) { result } if method
+    end
+  end
```
floehopper: All the same comments apply as for `any_instance_method_test.rb`.

lib/mocha/mockery.rb
```diff
@@ -194,6 +153,13 @@ def logger
 
     private
 
+    def check(action, description, object, method)
+      return if Mocha::Configuration.allow?(action) || (block_given? && !yield)
+      stubbing_error = "stubbing #{description}: #{object.mocha_inspect}.#{method}"
+      raise StubbingError.new(stubbing_error, caller) if Mocha::Configuration.prevent?(action)
+      logger.warn(stubbing_error) if Mocha::Configuration.warn_when?(action)
+    end
+
```
floehopper: I don't understand what this bit of the commit note means:

> I had to create method_exists? temporarily to please rubocop even though it stood out as different from all the other object.foo conditions.

As far as I can see this commit does not introduce `method_exists? `. So I don't understand what you mean by "I had to create method_exists? temporarily". Can you explain?

test/unit/class_method_test.rb
```diff
@@ -13,7 +13,7 @@ def class_with_method(method, result = nil)
     Class.new do
       extend ClassMethods
       singleton_class.extend(ClassMethods)
-      singleton_class.define_method(method) { result } if method
+      singleton_class.send(:define_method, method) { result } if method
```
floehopper: I'd prefer that this was included in the commit in which the original change was introduced.

test/acceptance/stubbing_non_existent_any_instance_method_test.rb
```diff
@@ -67,7 +67,7 @@ def test_should_allow_stubbing_method_to_which_any_instance_responds
     Mocha::Configuration.prevent(:stubbing_non_existent_method)
     klass = Class.new do
       def respond_to?(method, _include_private = false)
-        (method == :method_to_which_instance_responds)
+        (method.to_sym == :method_to_which_instance_responds)
```
floehopper: I'm not 100% sure I understand why these tests (and the other ones in this commit) have to change. Can you explain a bit more?

In any case I'd prefer if all the changes to the tests in this commit were incorporated into the commit which actually required them.

test/unit/class_method_test.rb
```diff
@@ -268,7 +268,7 @@ def test_should_match_if_other_class_method_has_same_stubbee_and_same_method_but
       def equal?(_other)
         false
       end
-    end
+    end.new
```
floehopper: Please fix up the relevant commit including this change. Thanks.

nitishr:
I've redone the whole branch. I think you'd find the commit history cleaner and more logical now. Could you please review?

floehopper:
> I've redone the whole branch. I think you'd find the commit history cleaner and more logical now. Could you please review? Hopefully, all your questions are answered through the commit sequence and commit notes.

Thanks so much. I'm going to resolve all my previous review comments and re-review the whole branch.

lib/mocha/object_methods.rb
```diff
@@ -163,13 +163,9 @@ def unstub(*method_names)
...
     def method_exists?(method, include_public_methods = true)
-      if include_public_methods
-        return true if public_methods(true).include?(method)
-        return true if respond_to?(method.to_sym)
-      end
-      return true if protected_methods(true).include?(method)
-      return true if private_methods(true).include?(method)
-      false
+      klass = singleton_class
+      (include_public_methods && (klass.method_defined?(method) || respond_to?(method.to_sym))) ||
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/class_methods.rb
```diff
@@ -31,13 +31,7 @@ def stubba_method
...
 
       def method_exists?(method, include_public_methods = true)
-        if include_public_methods
-          return true if @stubba_object.public_instance_methods(true).include?(method)
-          return true if @stubba_object.allocate.respond_to?(method.to_sym)
-        end
-        return true if @stubba_object.protected_instance_methods(true).include?(method)
-        return true if @stubba_object.private_instance_methods(true).include?(method)
-        false
+        stubba_object.allocate.method_exists?(method, include_public_methods)
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/object_methods.rb
```diff
@@ -164,7 +164,7 @@ def unstub(*method_names)
     # @private
     def method_exists?(method, include_public_methods = true)
       klass = singleton_class
-      (include_public_methods && (klass.method_defined?(method) || respond_to?(method.to_sym))) ||
+      (include_public_methods && klass.method_defined?(method)) ||
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/mockery.rb
```diff
@@ -194,6 +194,10 @@ def logger
 
     private
 
+    def method_exists?(object, method)
+      object.method_exists?(method, true) || object.respond_to?(method)
+    end
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/class_methods.rb
```diff
@@ -30,6 +30,10 @@ def stubba_method
 
       attr_reader :stubba_object
 
+      def respond_to?(method)
+        stubba_object.allocate.respond_to?(method)
+      end
+
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/class_methods.rb
```diff
@@ -56,9 +56,10 @@ def any_instance
     end
 
     # @private
-    def method_exists?(method, include_public_methods = true)
-      (include_public_methods && method_defined?(method)) ||
-        (protected_method_defined?(method) || private_method_defined?(method))
+    def method_visibility(method, include_public_methods = true)
+      (include_public_methods && public_method_defined?(method) && :public) ||
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/class_method.rb
```diff
@@ -91,13 +91,6 @@ def to_s
       "#{stubbee}.#{method_name}"
     end
 
-    def method_visibility
-      (original_method_owner.public_method_defined?(method_name) && :public) ||
-        (original_method_owner.protected_method_defined?(method_name) && :protected) ||
-        (original_method_owner.private_method_defined?(method_name) && :private)
-    end
-    alias_method :method_defined_in_stubbee_or_in_ancestor_chain?, :method_visibility
```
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

test/unit/any_instance_method_test.rb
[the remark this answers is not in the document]
floehopper: Resolving - the whole branch has been re-written.

test/unit/any_instance_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

test/unit/any_instance_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

test/unit/class_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/mockery.rb
floehopper: Resolving - the whole branch has been re-written.

test/acceptance/stubbing_non_existent_any_instance_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

test/unit/class_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

test/unit/class_method_test.rb
floehopper: Resolving - the whole branch has been re-written.

lib/mocha/mockery.rb
```diff
@@ -195,7 +195,7 @@ def logger
     private
 
     def method_exists_or_responds_to?(method, object)
-      object.method_exists?(method, true)
+      object.method_exists?(method, true) || object.respond_to?(method.to_sym)
```
floehopper: Can you explain how you determined that this is the only call site for `method_exists?` which needs the call to `respond_to?`?

nitishr: Because respond_to? was called only in the `if include_public_methods` conditional, and this is the only call site that sets `include_public_methods` to true.

nitishr: The only other call site is in the same file when we check for stubbing non-public methods. I found that by doing both a method reference search and a textual search.

nitishr:
Sorry, just force-pushed because I noticed I'd mistakenly removed `require 'method_definer'` from the test files. Putting it back [here](https://github.com/freerange/mocha/pull/370/commits/941c6ffcd3b421eae70a36560050c0cfaa1cd44f) was the only change in the force-push.

lib/mocha/mockery.rb
floehopper: Ah, of course! That makes sense. Could you add that explanation to the commit note?

nitishr: Done

floehopper:
> I have one other slight concern. Can you reassure me that it's definitely safe to implement a `singleton_class` method given that [this method is already defined on `Object`](https://ruby-doc.org/core-2.6.3/Object.html#method-i-singleton_class)? I wouldn't want us to mess with the definition of this on objects outside Mocha.

I've looked into this a bit and I now understand that you've only defined an alternative implementation of `Object#singleton_class` on `Mocha::ClassMethods::AnyInstance`, so I think it's pretty safe. Having said that I think it might still be preferable to define it as an alternative method or at least somehow retain the original implementation rather than over-writing it.

Anyway, I'm happy to merge this as it stands and address the above separately. I plan to rebase this against `master` and get it merged this afternoon (UK time).

floehopper:
I've rebased the work in this PR against `master` into [this tracking branch](https://github.com/freerange/mocha/compare/method_exists) on this repo. If/when the CI build passes, I'll get the changes merged into `master`.

floehopper:
Merged in 068d6f743d0010b153c10d92ae5b644fc27677fe. Closing.

Issue #353, opened 2019-03-09
Deprecate / prevent calling Mocha::API#mock or #stub with single symbol


floehopper:
See #347.

nitishr:
Should this be closed now?

floehopper:
Yes - good point!

This was addressed in #377, although instead of deprecating the undocumented behaviour, we now allow the name of a mock object to be a symbol or a string.

Closing.

PR #377, opened 2019-10-05
accept symbol as mock/stub name

nitishr:
Looking at #353 and #347 , I thought instead of deprecating / preventing calling Mocha::API#mock or #stub with single symbol, we could treat the first symbol as the test double name, i.e. accept strings as well as symbols for test double names.

There's a bit of duplication I'd like to remove, but only if the basic idea is acceptable.

### DIFF
--- a/lib/mocha/api.rb
+++ b/lib/mocha/api.rb
@@ -21,7 +21,7 @@ module Mocha
 
     # Builds a new mock object
     #
-    # @param [String] name identifies mock object in error messages.
+    # @param [String, Symbol] name identifies mock object in error messages.
     # @param [Hash] expected_methods_vs_return_values expected method name symbols as keys and corresponding return values as values - these expectations are setup as if {Mock#expects} were called multiple times.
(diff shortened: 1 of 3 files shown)


floehopper:
@nitishr I think this is a good idea. Thanks for suggesting it!

* I think I'd prefer the extra tests to be written out explicitly rather than using loops inside existing tests, even if that results in some duplication - that way the test name can be more specific.

* Although it doesn't look like it causes any problems, I think I'd prefer it if the `@name` instance variable in `Mocha::Name` was always converted to be a `String`.

I'll leave you in peace to remove the duplication you mentioned. Let me know when it's ready for a full review.

nitishr:
@floehopper ready for a full review (as are my other 2 PRs).

floehopper:
> ready for a full review (as are my other 2 PRs).

I've reviewed one of your other PRs, but unfortunately I've run out of time now. I'll try to review the remaining one as soon as I can. Thanks again for all your work! 😄

nitishr:
Done! Can you please re-review?

PR #372, opened 2019-10-05
Reduce duplication in AnyInstanceMethod class

nitishr:
As suggested in #359.

Also addresses #332

Build is currently failing due to minitest latest version not supporting ruby 1.8.6 as described [here](https://github.com/freerange/mocha/pull/370#issuecomment-537675002)

Issue #359, opened 2019-06-08
Reduce duplication in AnyInstanceMethod class


floehopper:
From #358:

> There are still some overridden methods in `AnyInstanceMethod` (notably `restore_original_method` and `stub_method_definition`) that it'd be good to investigate in future to see whether we can remove more duplication.

floehopper:
@nitishr has done a great job of removing all the duplication in #378. Closing.

Issue #332, opened 2018-05-15
Consider whether we need different implementations of stubbed_method_implementation in ClassMethod and AnyInstanceMethod


floehopper:
Extracted from https://github.com/freerange/mocha/pull/269#issuecomment-389104469.

> Do we need different implementations of `stubbed_method_implementation` (in `ClassMethod` and `AnyInstanceMethod`)?

floehopper:
@nitishr has addressed this in #378. Closing.

PR #378, opened 2019-10-06
Reduce duplication in any instance method class

nitishr:
Identical to https://github.com/freerange/mocha/tree/reduce-duplication-in-any-instance-method-class, except that it fixes the (newly added) hanging test by not snapshotting.

The test seemed to be hanging inside Snapshot constructor (`unbound_method = receiver.instance_method(method)`). Most other tests in the file don't check the snapshot, either. The test should be good enough without a snapshot.

This has removed the co-authoring because of a fixup & rebase. I'll try to fix that if I can figure out how to.

### DIFF
--- a/lib/mocha/any_instance_method.rb
+++ b/lib/mocha/any_instance_method.rb
@@ -15,8 +15,9 @@ module Mocha
       return if use_prepended_module_for_stub_method?
       if stub_method_overwrites_original_method?
         original_method_owner.send(:define_method, method_name, original_method)
-        Module.instance_method(original_visibility).bind(original_method_owner).call(method_name)
       end
+      return unless original_visibility
+      Module.instance_method(original_visibility).bind(original_method_owner).call(method_name)
     end
 
     private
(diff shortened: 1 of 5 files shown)


floehopper:
> The test seemed to be hanging inside Snapshot constructor (`unbound_method = receiver.instance_method(method)`).

Oh - interesting! It sounds similar to [a Ruby bug][1] I reported a couple of years ago!

[1]: https://bugs.ruby-lang.org/issues/12832

nitishr:
Yes, looks similar, maybe related. I noticed that `receiver.instance_methods` (where receiver is the module with private method) includes the method I'm expecting (`my_included_method`). However, `receiver.instance_method(:my_included_method)` hangs.

Anyway, I've updated all the commits to include you as co-author, and the build has passed now.

floehopper:
Oh, thanks for making me co-author - I only did that for the new commits I separated out, because otherwise I would've been the *only* author which didn't seem fair! I would've been happy for all the commits to have you as the only author.

> Most other tests in the file don't check the snapshot, either. The test should be good enough without a snapshot.

Agreed. I'll get it merged now.

floehopper:
As per the description of #372, this work was suggested in #359 and also addresses #332.

Issue #230, opened 2015-06-20
Allow passing a block to Expectation#returns


Hubro:
Instead of this:

```
class SomethingTest
  def setup
    @subject = Something.new
    @test_data = fetch_test_data
  end

  def test_foo
    @subject.stub(:fetch_over_network).returns(test_data["foo"])
    # ...
  end

  def test_bar
    @subject.stub(:fetch_over_network).returns(test_data["bar"])
    # ...
  end

  def test_baz
    @subject.stub(:fetch_over_network).returns(test_data["baz"])
    # ...
  end
end
```

I would like to be able to do this:

```
class SomethingTest
  def setup
    @subject = Something.new
    @subject.stub(:fetch_over_network).returns { |input| test_data[input] }
  end

  # ...
end
```

I yearn to be able to just pass a block to `returns` and decide the return value based on the input arguments. It would save so much boilerplate in situations like this...

Additionally, it happens too often that I forget to stub a method, or I write a typo (`:fetch_over_netrwok`) and end up actually fetching something over the network every time my test suite runs.

To summarize, I feel like implementing this would be very DRY of you, and would in some situations drastically reduce the amount of boilerplate code in tests.

booleanbetrayal:
:+1:

floehopper:
Closing in favour of #224.

Issue #213, opened 2015-02-10
Returns with block and inspecting arguments?


hakunin:
Hi, I am moving away from now not maintained gem `rr` and wanted to ask if there is any plan to be able to do this (pseudocode):

```
args = []
SomeClass.any_instance.stub(:perform) do  |a, b, c|
  args << [a, b, c]; nil
end

assert_equal "what I expected to get", args
```

I see that the current code could be extended to do that, would there be interest for a pull request?

floehopper:
@hakunin: I'm busy with work at the moment, but I'll reply as soon as I can.

floehopper:
@hakunin: I'm sorry not to have replied for ages.

Can you explain a bit more about exactly what you're trying to achieve? You can achieve something pretty similar to what it looks like you want as follows:

```ruby
class FooTest < Minitest::Test
  class Foo
    def self.bar(*args)
    end
  end

  def test_foo
    Foo.expects(:bar).with(1, 2, 3)
    Foo.bar(2, 3, 4)
  end
end
```

This gives the following output:

```
$ ruby -Itest foo.rb
Run options: --seed 62486

# Running:

F

Finished in 0.002787s, 358.8088 runs/s, 0.0000 assertions/s.

  1) Failure:
FooTest#test_foo [foo.rb:13]:
unexpected invocation: FooTest::Foo.bar(2, 3, 4)
unsatisfied expectations:
- expected exactly once, not yet invoked: FooTest::Foo.bar(1, 2, 3)


1 runs, 0 assertions, 1 failures, 0 errors, 0 skips
```

i.e. this output shows what arguments were expected and what arguments were actually received.

hakunin:
Hi, thanks for the reply. My goals are:

1) sometimes I just need to run code instead of the method
2) one use-case is to collect the arguments, the method is called multiple times though

Hubro:
It would be useful to be able to stun out a method and replace it with another method. Currently I am doing this to stub out methods that go over the network:

def setup
  data = get_test_data("addresses")

  @subject = ClassToBeTested.new

  @subject.define_singleton_method(:a_search_address) { |address|
    data[:a][address] or fail "Unexpected address: #{address}"
  }

  @subject.define_singleton_method(:b_search_address) { |address|
    data[:b][address] or fail "Unexpected address: #{address}"
  }

  @subject.define_singleton_method(:c_search_address) { |address|
    data[:c][address] or fail "Unexpected address: #{address}"
  }
end

[...]

PR #379, opened 2019-10-07
Allow responding with result of a block

nitishr:
Inspired by the suggestion from #230. 

I took the approach of adding a different expectation methods (`responds`) rather than extend `returns`, because:

 - Allowing either *values or a block for `returns` seemed like an unnecessary restriction when you could allow both at the same time, but
 - Allowing both *values and a block seemed a little unintuitive/unclear in intent (e.g. `returns(1, 2) { |n| n * 2 }`)

### DIFF
--- a/lib/mocha/exception_raiser.rb
+++ b/lib/mocha/exception_raiser.rb
@@ -5,7 +5,7 @@ module Mocha
       @message = message
     end
 
-    def evaluate
+    def evaluate(*_args)
       raise @exception, @exception.to_s if @exception.is_a?(Module) && (@exception < Interrupt)
       raise @exception, @message if @message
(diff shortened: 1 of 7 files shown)


nitishr:
Looks like #231 was pretty much the same as/very similar to the solution proposed here, but was declined. There're a number of issues requesting the same/similar feature still open. It'd be useful to close those: e.g. #230, #213.

floehopper:
@nitishr Thanks for your work on this. As you've mentioned elsewhere there have been a number of requests along these lines. I'd prefer to have some further discussion about the purpose of this feature and how best to implement it and so I'm going to close this PR in favour of continuing discussion on #224. I hope that's OK.

Issue #280, opened 2016-10-28
Add some acceptance tests for stubbing methods defined in refinements


floehopper:
Refinements have been around for a while now, but I don't think we currently have any tests to check that stubbing works for such methods.

PR #380, opened 2019-10-07
Add acceptance tests for stubbing methods defined in refinements

nitishr:
As suggested in #280


nitishr:
Looking at the failing builds, there seem to be too many inconsistencies between refinements in different ruby versions for this to be worth it, IMO.

PR #381, opened 2019-10-09
Simplify {AnyInstance,Class,Instance,Module}Methods Hierarchy

nitishr:
It appears to me that there're essentially only two types of method stubbing:
a) on a non-specific object/instance (i.e. on `any_instance` of a class)
b) on a specific object/instance (where this specific object might also be a class or a module - it's ruby, after all)

For (b), it makes no difference (at least in the current implementation) whether it's a ClassMethod, ModuleMethod, or InstanceMethod - which is why the latter were empty subclasses of ClassMethod. I think it's simpler to think in terms of specific & non-specific instances, rather than list out all forms of specific instances even when there're no differences between them.

This is why I've gotten rid of ModuleMethod and ClassMethod (even if they did exist, it'd be more appropriate for them to inherit from InstanceMethod than for InstanceMethod and ModuleMethod to inherit from ClassMethod). Instead, I've introduced the abstract notion of a StubbedMethod. I'm not entirely sure if I've quite nailed the name or the concept right, although I feel fairly comfortable with that. Alternatively, I was thinking of it as a `StubbingMethod` (or `StubbingStrategy`).

This change has also allowed me to shrink the contract required _by_ the subclass _from_ the superclass down to just one method - `stubbee`, and to make the subclass responsibilities totally effect-free.


lib/mocha/any_instance_method.rb
```diff
@@ -14,7 +14,7 @@ def original_method_body
     end
 
     def store_original_method
-      @original_method = original_method_owner.instance_method(method_name)
+      @original_method = stubbee.instance_method(method_name)
```
floehopper: Can you explain how this is definitely equivalent? i.e. how you know there is no change in behaviour?

test/unit/instance_method_test.rb
```diff
@@ -226,35 +225,35 @@ def test_should_return_mock_for_stubbee
...
     stubbee = Object.new
     stubbee.define_instance_method(:mocha) { mocha }
-    method = ClassMethod.new(stubbee, :method_name)
+    method = InstanceMethod.new(stubbee, :method_name)
     assert_equal mocha, method.mock
   end
 
   def test_should_not_match_if_other_object_has_a_different_class
-    class_method = ClassMethod.new(Object.new, :method)
+    class_method = InstanceMethod.new(Object.new, :method)
```
floehopper: I think there are a bunch of places like this where the local variable, `class_method` should be renamed to `instance_method` or perhaps just `method`.

lib/mocha/any_instance_method.rb
nitishr: This is definitely equivalent because I've essentially inlined the `original_method_owner` call here, i.e. `original_method_owner` returns `stubbee`.

test/unit/instance_method_test.rb
nitishr: Fixed now

nitishr:
> I think StubbingStrategy is better than StubbedMethod, but that probably means also renaming ClassMethod -> ClassMethodStubbingStrategy and AnyInstanceMethod -> AnyInstanceMethodStubbingStrategy. What do you think?

Having looked at the client code, I feel more comfortable with `StubbedMethod` now (even though more correctly, it should be something like `MethodToStub`/`MethodToBeStubbed`. See, for instance: https://github.com/freerange/mocha/blob/master/lib/mocha/object_methods.rb#L29. Would I rename the method to `stubba_strategy`/`stubbing_strategy`? What about https://github.com/freerange/mocha/blob/master/lib/mocha/object_methods.rb#L79-L80? Replacing method with strategy doesn't work there either. Also, generally, I think it's better to name classes after a real thing rather than a process/procedure dressed up as a thing (Strategy is a (representation of a) process as it describes _how_ to do something).

Looking at some old commit notes as well, which I can't dig out right now, it seems the original thinking was for those classes to represent the method being stubbed, rather than the method _of_ stubbing (aka strategy).

Thoughts?

floehopper:
> Having looked at the client code, I feel more comfortable with `StubbedMethod` now (even though more correctly, it should be something like `MethodToStub`/`MethodToBeStubbed`. See, for instance: https://github.com/freerange/mocha/blob/master/lib/mocha/object_methods.rb#L29. Would I rename the method to `stubba_strategy`/`stubbing_strategy`? What about https://github.com/freerange/mocha/blob/master/lib/mocha/object_methods.rb#L79-L80? Replacing method with strategy doesn't work there either. Also, generally, I think it's better to name classes after a real thing rather than a process/procedure dressed up as a thing (Strategy is a (representation of a) process as it describes _how_ to do something).
> 
> Looking at some old commit notes (from @floehopper) as well, which I can't dig out right now, it seems the original thinking was for those classes to represent the method being stubbed, rather than the method _of_ stubbing (aka strategy).
> 
> Thoughts?

I think what you say all sounds reasonable.

For additional context, the word `stubba` that appears in a few places in the codebase is a bit of a vestige of a bygone era in the project. However, one advantage of `Mocha::ObjectMethods#stubba_method` is that it's an obscure method name that's unlikely to interfere with anyone else's code. But, of course, there are other ways to achieve that.

Thinking about it a bit more, I think something like `ClassMethodStubber` & `AnyInstanceMethodStubber` might be better.

Anyway, on reflection, I'd be happy to merge this PR without any further changes. We can always address the name change in a later PR. I'll approve this PR now and get it merged.

Issue #352, opened 2018-11-27
More realistic examples for Expectation#yields and Expectation#multiple_yields


floehopper:
In #351 @itmart didn't understand how `Expectation#yields` worked and in fact in some ways his question was more related to `Expectation#multiple_yields`. I'm not convinced the examples in the documentation are as useful as they could be.

floehopper:
Closed by #383.

PR #382, opened 2019-10-15
Make yields fail when the caller does not provide a block

nitishr:
This is more reasonable and realistic than yielding only if a block is given. A stub that's set up to yield must unconditionally yield, just like it would unconditionally raise or return when set up to do so. The block_given? check should be the responsibility of the original method implementation if it so chooses. A stub set up to yield is presumably set up to simulate the code path that actually yields.

In rspec-mocks as well, [It fails when the caller does not provide a block]( https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/configuring-responses/yielding#it-fails-when-the-caller-does-not-provide-a-block). (Note: Unlike rspec-mocks, I did not choose that [It fails when the caller's block does not accept the provided arguments](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/configuring-responses/yielding#it-fails-when-the-caller's-block-does-not-accept-the-provided-arguments) because that's not how yielding from a real method would behave.)

P.S. This change was prompted by looking at rspec-mocks docs to get ideas for #352.

### DIFF
--- a/test/unit/expectation_test.rb
+++ b/test/unit/expectation_test.rb
@@ -112,6 +112,11 @@ class ExpectationTest < Mocha::TestCase
     assert_equal [], yielded_parameters
   end
 
+  def test_yield_should_fail_when_the_caller_does_not_provide_a_block
+    expectation = new_expectation.yields(:foo)
+    assert_raise(LocalJumpError) { expectation.invoke }
+  end
+
   def test_should_yield_with_specified_parameters
     expectation = new_expectation.yields(1, 2, 3)
     yielded_parameters = nil
--- a/test/unit/expectation_test.rb
+++ b/test/unit/expectation_test.rb
@@ -17,18 +17,15 @@ class ExpectationTest < Mocha::TestCase
   end
 
   def test_should_match_calls_to_same_method_with_exactly_zero_parameters
-    expectation = new_expectation.with
-    assert expectation.match?(:expected_method)
+    assert new_expectation.with.match?(:expected_method)
   end
 
   def test_should_not_match_calls_to_same_method_with_more_than_zero_parameters
-    expectation = new_expectation.with
-    assert !expectation.match?(:expected_method, 1, 2, 3)
+    assert !new_expectation.with.match?(:expected_method, 1, 2, 3)
   end
 
   def test_should_match_calls_to_same_method_with_expected_parameter_values
-    expectation = new_expectation.with(1, 2, 3)
-    assert expectation.match?(:expected_method, 1, 2, 3)
+    assert new_expectation.with(1, 2, 3).match?(:expected_method, 1, 2, 3)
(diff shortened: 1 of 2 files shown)


test/unit/expectation_test.rb
```diff
@@ -112,11 +112,9 @@ def test_should_yield_no_parameters
     assert_equal [], yielded_parameters
   end
 
-  def test_should_yield_with_specified_parameters
```
floehopper: I don't understand why you've removed this test where a block *is* passed to the stubbed method. Surely it's still relevant...?

nitishr: Oops! Restored.

nitishr:
Updated the docs. I think deprecating would be a good idea since this change could potentially break some tests.

I'd be happy to make the necessary changes to deprecate if you could point me to how it's done in mocha.

floehopper:
> Updated the docs. I think deprecating would be a good idea since this change could potentially break some tests.

The documentation changes look good.

> I'd be happy to make the necessary changes to deprecate if you could point me to how it's done in mocha.

The mechanism is a bit antiquated, but if you search for usages of `Mocha::Deprecation.warning`, I think that should give you something to go on.

nitishr:
So, would that mean parking this PR, and opening another PR with something like
```
      if block_given?
        @yield_parameters.next_invocation.each do |yield_parameters|
          yield(*yield_parameters)
        end
      elsif @yield_parameters.any?
        Deprecation.warning('... message here ...')
      end
```

If yes, this would mean adding the `any?` method to `YieldParameters` only in order to support deprecation. Genuinely not sure if that effort for deprecating is justified. I'd suspect extremely low impact of this change, but I'd rely more on your opinion, given I have little clue about mocha's uses - never used it myself either.

floehopper:
I think you're probably right in this case. This PR effectively changes an _undocumented_ aspect of the public API's behaviour (to be more correct).

Am I right in thinking that if anybody's tests start failing when they upgrade Mocha all they'll need to do to fix it is remove the call to `yields` or `multiple_yields` which wasn't doing anything when they were using the previous version of Mocha...?

Mocha is [used in a lot of projects](https://rubygems.org/gems/mocha), so we *do* need to be careful about deprecations and semantic versioning when making changes.

nitishr:
> Am I right in thinking that if anybody's tests start failing when they upgrade Mocha all they'll need to do to fix it is remove the call to yields or multiple_yields which wasn't doing anything when they were using the previous version of Mocha...?

You are right. Or provide a block when invoking the method.

floehopper:
I'm going to rebase this against `master` and get it merged.

floehopper:
Merged in e9e117e6daf5bbf3e08ed7511ede4df2c6d612bf. Closing.

PR #383, opened 2019-10-15
More realistic examples for yields and multiple_yields

nitishr:
As suggested in #352


floehopper:
These examples look much better - thank you! I would merge this as it stands, but the build is failing. I contemplated kicking the build, but it looks like a bunch of builds are in progress, so I won't make things more complicated.

nitishr:
Surely a transient error due to a glitch on travis. I had some other branch builds failing for the same reason as well, and they passed on restarting the failed jobs. Could you please try restarting failed jobs/builds?

floehopper:
Closes #352.

Issue #330, opened 2018-04-18
Add documentation to explain how Mocha is intended to be used


floehopper:
I think it would be useful for the Mocha documentation to point out the different types of Test Double and that it's only intended to be used for the Mock Object or Test Stub types. The following from a comment in #317 might be useful:

> Mocha is intended to be used in *unit* tests for the [Mock Object][] or [Test Stub][] types of [Test Double][], not the [Fake Object][] or [Test Spy][] types. Although it would be possible to extend Mocha to allow the implementation of fakes and spies, I'd prefer to keep it focussed on mocks and stubs. I think providing a fully-fledged solution for fakes and/or spies would be a lot of work.

[Mock Object]: http://xunitpatterns.com/Mock%20Object.html
[Test Stub]: http://xunitpatterns.com/Test%20Stub.html
[Test Double]: http://xunitpatterns.com/Test%20Double.html
[Fake Object]: http://xunitpatterns.com/Fake%20Object.html
[Test Spy]: http://xunitpatterns.com/Test%20Spy.html

floehopper:
Closed in #385.

PR #385, opened 2019-10-15
Add documentation to explain how Mocha is intended to be used

nitishr:
It's useful for the Mocha documentation to point out the different types of Test
Double and that it's only intended to be used for the Mock Object or Test Stub
types, as suggested in #330

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,13 @@
 
 ### Description
 
-* A Ruby library for mocking and stubbing.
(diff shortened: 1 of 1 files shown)


floehopper:
I agree with this gist of this, but I'd like to make some tweaks to the copy when I have time. Sorry for the delay.

floehopper:
On reflection, I'm happy to merge this as it stands. Thanks again for your contribution.

Issue #278, opened 2016-10-27
Consider re-using existing ClassMethod::PrependedModule rather than always prepending a new instance


floehopper:
We currently [_always_ prepend](https://github.com/freerange/mocha/blob/master/lib/mocha/class_method.rb#L45) an instance of `ClassMethod::PrependedModule` even if an instance has already been prepended. While i don't think this is currently causing a problem, it feels messy and can be confusing when debugging. The only thing we might need to think about is what happens when you stub the same method twice on the same definition target.

PR #387, opened 2019-10-16
don't prepend if already prepended PrependedModule

nitishr:
We used to always prepend an instance of PrependedModule even if an instance has
already been prepended. While this wasn't causing a problem, it felt messy and
could be confusing when debugging.

As suggested in #278

### DIFF
--- a/lib/mocha/stubbed_method.rb
+++ b/lib/mocha/stubbed_method.rb
@@ -123,8 +123,10 @@ module Mocha
     end
 
     def use_prepended_module_for_stub_method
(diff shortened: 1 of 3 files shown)


nitishr:
Closing because ruby-2.0 doesn't include the singleton class in its ancestors list.

floehopper:
> Closing because ruby-2.0 doesn't include the singleton class in its ancestors list.

Does this mean that #278 isn't an issue for Ruby >= v2? Or have I misunderstood?

Issue #229, opened 2015-05-04
Fail fast when using mocha/test_unit or mocha/minitest if integration fails


floehopper:
If someone is explicitly requiring "mocha/test_unit" or "mocha/minitest" then the implication is that they know which test framework they are trying to integrate with. If the Mocha integration fails (because the test framework hasn't been loaded) then we should raise an error. But as a first step we should deprecate this usage to give people a chance to fix things.

floehopper:
This has been partially addressed in #389 / c6032d0 in that these usages have been deprecated, but I'm going to leave this issue open to remind us to fail fast in a later release.

PR #389, opened 2019-10-16
Warn if integration using mocha/{test_unit,minitest} fails

nitishr:
As suggested in #229

### DIFF
--- a/lib/mocha/minitest.rb
+++ b/lib/mocha/minitest.rb
@@ -1,3 +1,5 @@
 require 'mocha/integration/mini_test'
 
-Mocha::Integration::MiniTest.activate
+unless Mocha::Integration::MiniTest.activate
+  Deprecation.warning("MiniTest must be loaded *before* `require 'mocha/minitest'`.")
+end
--- a/lib/mocha/test_unit.rb
+++ b/lib/mocha/test_unit.rb
@@ -1,3 +1,5 @@
 require 'mocha/integration/test_unit'
 
-Mocha::Integration::TestUnit.activate
+unless Mocha::Integration::TestUnit.activate
+  Deprecation.warning("Test::Unit must be loaded *before* `require 'mocha/test_unit'`.")
+end


floehopper:
Unfortunately #229 didn't really explain the issue thoroughly enough. I've updated the description of that issue to explain why we're not yet ready to apply the changes in this PR.

floehopper:
Also I was imagining using Ruby's `Kernel#warn` directly since I don't think that strictly speaking this is a deprecation as such.

nitishr:
Would it help if I changed the message from:

`Deprecation.warning("MiniTest must be loaded *before* `require 'mocha/minitest'`.")`

to

`warn("MiniTest must be loaded *before* loading mocha.")`

for minitest and a similar change for test-unit, i.e. not including the specific `require` statement (`require mocha/minitest` or `require mocha/test_unit`) in the warning message explicitly?

I think it'll be less/not confusing then, and will be more helpful than silence for someone whose test library integration has failed.

Even otherwise, I think it's more helpful to warn someone for whom the integration has failed than to avoid what a _tiny_ (IMO) bit of confusion for someone who tried to integrate using the legacy/undocument `require mocha/setup` option.

floehopper:
I'm away at the moment, but I'll look at this as soon as I get time. Sorry for the delay.

floehopper:
I've belatedly realised that I was confused in my previous comments on this PR. I had confused `mocha/test_unit.rb` and `mocha/minitest.rb` with `mocha/integration/test_unit.rb` and `mocha/integration/mini_test.rb`. The former are *not* loaded/called by `mocha/setup.rb` whereas the latter are. I've updated #229 to reflect this (i.e. pretty much reverted it back to how it was). I've also created a new issue (#398) as a reminder to deprecate the use of "mocha/setup" which I now understand is a separate issue.

I also agree that it's sensible to use deprecation warnings in this PR as a first step; then as a later step we can fail fast. I don't know what I was thinking when I suggested using `Kernel#warn`.

Sorry for all the confusion. I will get this merged shortly.

floehopper:
Tweaked very slightly and merged in c6032d0bbfe6bb64295f9d98e94ff9814085154e. Closing.

Issue #207, opened 2014-06-19
Improve documentation around test framework integration


floehopper:
See [this comment](https://github.com/freerange/mocha/issues/193#issuecomment-36508036).

PR #390, opened 2019-10-16
Improve documentation around test framework integration

nitishr:
As suggested in #207

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,50 +16,58 @@ Install the latest version of the gem with the following command...
 
     $ gem install mocha
 
-Note: If you are intending to use Mocha with Test::Unit or MiniTest, you should only setup Mocha *after* loading the relevant test library...
-
-##### Test::Unit
-
+And activate the Mocha gem in the test file(s) with:
 ```ruby
 require 'rubygems'
 gem 'mocha'
+```
+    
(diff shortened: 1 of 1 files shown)


floehopper:
When I first looked at this I found it hard to see what had changed and to understand why it had changed. I'll have a closer look at it when I have time, but if you can break it up a bit more and explain the motivation behind each change, that would really help.

floehopper:
Sorry not to have got back to you about this for ages. I've had another look at it and I now think I understand the changes you've made.

It's worth noting that #207 was opened in the context of a very specific problem, i.e. that at one point the Test::Unit in Ruby's standard library was at one point a thin wrapper around MiniTest. I believe this was intended to ease the transition from Test::Unit to MiniTest which at that point was then included in the standard library. Since then both Test::Unit and MiniTest have been removed from the standard library and so I think this specific issue is less likely to be problematic other than for old projects.

I understand why you've split out a "Setup" section from the "Installation" section, but I find the result a bit hard to follow. Also the auto-detection mechanism provided by `require 'mocha/setup'` has been deprecated, although probably only since you opened this PR.

I'm going to have a look at the README in the light of #207 and this PR and see whether I think any changes are worthwhile.

Issue #391, opened 2019-10-18
Don't over-write definition of singleton_class method on Mocha::ClassMethods::AnyInstance


floehopper:
See [this comment](https://github.com/freerange/mocha/pull/370#issuecomment-543689925) for details.

floehopper:
Closed by #392.

PR #392, opened 2019-10-18
Don't over-write definition of singleton_class

nitishr:
Introduce stubba_class method as part of the contract between Mockery and the
object being stubbed.

As suggested in #391


floehopper:
This looks good to me. Many thanks! 😄

Issue #178, opened 2014-01-05
Display matching invocations alongside their respective expectations in error message


floehopper:
See #167. Display all the actual invocations recorded against both unsatisfied & satisfied expectations, including the parameter values passed into the method and the return value generated by the expectation. I'm thinking along the following lines:

``` ruby
foo = mock('foo')
foo.expects(:bar).with(1).returns('a')
foo.stubs(:bar).with(2).returns('b').then.returns('c').then.raises(StandardError)

3.times { foo.bar(2) }
foo.bar(3)
```

```
unexpected invocation: #<Mock:foo>.bar(3)
unsatisfied expectations:
- expected exactly once, not yet invoked: #<Mock:foo>.bar(1)
satisfied expectations:
- allowed any number of times, invoked 3 times: #<Mock:foo>.bar(2)
  - #<Mock:foo>.bar(2) # => 'b'
  - #<Mock:foo>.bar(2) # => 'c'
  - #<Mock:foo>.bar(2) # => raised StandardError
```

It would also be nice to record any yields that occur.

We might only want to enable this using a verbose configuration option.

floehopper:
Implemented in #394 & #410.

floehopper:
See also https://github.com/freerange/mocha/issues/632.

PR #394, opened 2019-10-24
Display matching invocations alongside expectations

nitishr:
As suggested in #178

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -593,12 +593,7 @@ module Mocha
     # @private
     def mocha_inspect
       message = "#{@cardinality.mocha_inspect}, "
-      message << case @invocations.size
-                 when 0 then 'not yet invoked'
-                 when 1 then 'invoked once'
-                 when 2 then 'invoked twice'
-                 else "invoked #{@invocations.size} times"
-                 end
+      message << actual_invocations(@invocations.size)
       message << ': '
       message << method_signature
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
@@ -612,6 +607,19 @@ module Mocha
     end
 
     private
+
+    def actual_invocations(invocation_count)
+      case invocation_count
+      when 0 then
+        'not yet invoked'
+      when 1 then
+        'invoked once'
+      when 2 then
+        'invoked twice'
+      else
+        "invoked #{invocation_count} times"
+      end
+    end
 
     def method_name
       "#{@mock.mocha_inspect}.#{@method_matcher.mocha_inspect}"
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -555,12 +555,12 @@ module Mocha
 
     # @private
     def invocations_allowed?
-      @cardinality.invocations_allowed?(@cardinality.invocations.size)
+      @cardinality.invocations_allowed?
     end
 
     # @private
     def satisfied?
-      @cardinality.satisfied?(@cardinality.invocations.size)
+      @cardinality.satisfied?
     end
 
     # @private
@@ -574,12 +574,12 @@ module Mocha
     # @private
     def verified?(assertion_counter = nil)
       assertion_counter.increment if assertion_counter && @cardinality.needs_verifying?
-      @cardinality.verified?(@cardinality.invocations.size)
+      @cardinality.verified?
     end
 
     # @private
     def used?
-      @cardinality.used?(@cardinality.invocations.size)
+      @cardinality.used?
     end
 
     # @private
@@ -592,7 +592,7 @@ module Mocha
     # @private
     def mocha_inspect
       message = "#{@cardinality.mocha_inspect}, "
-      message << @cardinality.actual_invocations(@cardinality.invocations.size)
+      message << @cardinality.actual_invocations
       message << ': '
       message << method_signature
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -516,6 +516,7 @@ module Mocha
       @return_values = ReturnValues.new
       @yield_parameters = YieldParameters.new
       @backtrace = backtrace || caller
+      @returned_values = []
     end
 
     # @private
@@ -570,7 +571,9 @@ module Mocha
       @yield_parameters.next_invocation.each do |yield_parameters|
         yield(*yield_parameters)
       end
-      @return_values.next
+      returned_value = @return_values.next
+      @returned_values << returned_value
+      returned_value
     end
 
     # @private
@@ -603,12 +606,19 @@ module Mocha
       message << ': '
       message << method_signature
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
+      message << invocations if (ENV['MOCHA_OPTIONS'] || '').split(',').include?('verbose')
       message
     end
 
     # @private
     def method_signature
       "#{@mock.mocha_inspect}.#{@method_matcher.mocha_inspect}#{@parameters_matcher.mocha_inspect}"
+    end
+
+    private
+
+    def invocations
+      @returned_values.map { |returned_value| "\n  - #{method_signature} # => #{returned_value.mocha_inspect}"}.join
     end
   end
 end
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -591,10 +591,7 @@ module Mocha
 
     # @private
     def mocha_inspect
-      message = "#{@cardinality.anticipated_times}, "
-      message << @cardinality.invoked_times
-      message << ': '
-      message << method_signature
+      message = "#{@cardinality.anticipated_times}, #{@cardinality.invoked_times}: #{method_signature}"
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
       message << @cardinality.actual_invocations if (ENV['MOCHA_OPTIONS'] || '').split(',').include?('verbose')
       message
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -516,7 +516,6 @@ module Mocha
       @return_values = ReturnValues.new
       @yield_parameters = YieldParameters.new
       @backtrace = backtrace || caller
-      @invocations = []
     end
 
     # @private
@@ -556,12 +555,12 @@ module Mocha
 
     # @private
     def invocations_allowed?
-      @cardinality.invocations_allowed?(@invocations.size)
+      @cardinality.invocations_allowed?(@cardinality.invocations.size)
     end
 
     # @private
     def satisfied?
-      @cardinality.satisfied?(@invocations.size)
+      @cardinality.satisfied?(@cardinality.invocations.size)
     end
 
     # @private
@@ -575,12 +574,12 @@ module Mocha
     # @private
     def verified?(assertion_counter = nil)
       assertion_counter.increment if assertion_counter && @cardinality.needs_verifying?
-      @cardinality.verified?(@invocations.size)
+      @cardinality.verified?(@cardinality.invocations.size)
     end
 
     # @private
     def used?
-      @cardinality.used?(@invocations.size)
+      @cardinality.used?(@cardinality.invocations.size)
     end
 
     # @private
@@ -593,7 +592,7 @@ module Mocha
     # @private
     def mocha_inspect
       message = "#{@cardinality.mocha_inspect}, "
-      message << @cardinality.actual_invocations(@invocations.size)
+      message << @cardinality.actual_invocations(@cardinality.invocations.size)
       message << ': '
       message << method_signature
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
@@ -613,7 +612,7 @@ module Mocha
     end
 
     def invocations
-      @invocations.map(&:mocha_inspect).join
+      @cardinality.invocations.map(&:mocha_inspect).join
     end
   end
 end
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -569,12 +569,9 @@ module Mocha
     def invoke
       @invocation_count += 1
       perform_side_effects
-      invocation = Invocation.new(@return_values)
+      invocation = Invocation.new(@yield_parameters, @return_values)
       @invocations << invocation
-      @yield_parameters.next_invocation.each do |yield_parameters|
-        yield(*yield_parameters)
-      end
-      invocation.call
+      invocation.call { |*args| yield(*args) }
     end
 
     # @private
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -513,7 +513,6 @@ module Mocha
       @ordering_constraints = []
       @side_effects = []
       @cardinality = Cardinality.exactly(1)
-      @invocation_count = 0
       @return_values = ReturnValues.new
       @yield_parameters = YieldParameters.new
       @backtrace = backtrace || caller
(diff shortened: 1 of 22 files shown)


floehopper:
At first glance, this looks really promising. I'll try to do a full review as soon as I can, but it might not be for a couple of days. Thanks again for all your contributions - they are much appreciated! 😄

lib/mocha/mock.rb
```diff
@@ -305,11 +305,11 @@ def method_missing(symbol, *arguments, &block)
         raise NoMethodError, "undefined method `#{symbol}' for #{mocha_inspect} which responds like #{@responder.mocha_inspect}"
       end
       if (matching_expectation_allowing_invocation = all_expectations.match_allowing_invocation(symbol, *arguments))
-        matching_expectation_allowing_invocation.invoke(&block)
+        matching_expectation_allowing_invocation.invoke(*arguments, &block)
```
floehopper: As an aside, I like the fact that the invocation is now encapsulated in an object and I wonder if it might now make sense for methods like `ExpectationList#match_allowing_invocation`, `ExpectationList#match` & `ExpectationList#matches_method?` to take a invocation object as their argument...? Also I wonder if there is now some duplication between `Invocation` and `UnexpectedInvocation`...? I've captured these thoughts in https://github.com/freerange/mocha/issues/408 and https://github.com/freerange/mocha/issues/409.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -15,88 +15,45 @@ def teardown
...
+  def test_should_display_results_of_matching_invocations_alongside_expectations_in_verbose_mode
     test_result = run_as_test do
       foo = mock('foo')
       foo.expects(:bar).with(1).returns('a')
-      foo.stubs(:bar).with(2).returns('b').then.returns('c')
+      foo.stubs(:bar).with(any_parameters).returns('f').raises(StandardError)
 
-      2.times { foo.bar(2) }
-      foo.bar(3)
+      foo.bar(1, 2) { |_ignored| }
```
floehopper: Minor: I'm guessing the underscore prefix was added here (and in other lines in this test) to fix a Rubocop violation. It would be nice if this was done in the first commit where it was needed.

lib/mocha/expectation.rb
```diff
@@ -557,32 +556,31 @@ def match?(actual_method_name, *actual_parameters)
...
     end
 
     # @private
     def invoke(*arguments)
-      @invocation_count += 1
       perform_side_effects
       invocation = Invocation.new(method_name, @yield_parameters, @return_values)
       @invocations << invocation
-      invocation.call(*arguments) { |*args| yield(*args) }
+      invocation.call(*arguments) { |*yield_args| yield(*yield_args) }
```
floehopper: Minor: The `args` -> `yield_args` rename appears to be orthogonal to the stated purpose of this commit. It would be good to extract it into its own commit.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -1,7 +1,7 @@
 require File.expand_path('../acceptance_test_helper', __FILE__)
 require 'mocha/setup'
 
-class DisplayMatchingInvocationsAlongsideExpectations < Mocha::TestCase
+class DisplayMatchingInvocationsAlongsideExpectationsInVerboseMode < Mocha::TestCase
```
floehopper: Minor: The renaming of the test class and its methods is orthogonal to the stated purpose of this commit ("add test for yielding & returning nothing") and so it would be good to extract it into its own commit. Also it would be good to rename the file to match the class name.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -56,4 +56,23 @@ def test_should_display_yields_of_matching_invocations_alongside_expectations_in
       '  - #<Mock:foo>.bar(3, 4) # => raised StandardError after yielding ("b", "c"), then ("d", "e")'
     ], test_result.failure_message_lines
   end
+
+  def test_should_display_empty_yield_and_return
```
floehopper: Minor: I assume you've added this test just for completeness/coverage and not because it identified a problem with the implementation, i.e. I assume it would have passed even if you'd added it earlier. Is that correct? If so, it would be good to amend the earliest possible commit to include it. If on the other hand this test drove out some implementation changes, it would be good to include it in the commit where those changes were made.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -24,16 +24,12 @@ def test_should_display_results
       foo.bar(1, 2) { |_ignored| }
       assert_raise(StandardError) { foo.bar(3, 4) { |_ignored| } }
     end
-    assert_failed(test_result)
-    assert_equal [
-      'not all expectations were satisfied',
-      'unsatisfied expectations:',
-      '- expected exactly once, not yet invoked: #<Mock:foo>.bar(1)',
-      'satisfied expectations:',
+    assert_invocations(
```
floehopper: Minor: It would be helpful to expand the commit note ("DRY up tests") a bit to explain which tests and possibly mention extracting the `assert_invocations` method.

floehopper:
Just to expand slightly on my review, I think the most pressing issues are handling the `Expectaction#throws` scenario and avoiding rescuing `Exception`.

lib/mocha/expectation.rb
```diff
@@ -603,12 +605,19 @@ def mocha_inspect
       message << ': '
       message << method_signature
       message << "; #{@ordering_constraints.map(&:mocha_inspect).join('; ')}" unless @ordering_constraints.empty?
+      message << invocations if (ENV['MOCHA_OPTIONS'] || '').split(',').include?('verbose')
```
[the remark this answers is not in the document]
floehopper: That's fine. I'm happy for this PR to include that duplication. I have opened a new issue (https://github.com/freerange/mocha/issues/407) to capture the idea of refactoring the duplication.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -0,0 +1,38 @@
+require File.expand_path('../acceptance_test_helper', __FILE__)
+require 'mocha/setup'
+
+class DisplayMatchingInvocationsAlongsideExpectations < Mocha::TestCase
+  include AcceptanceTest
+
+  def setup
+    setup_acceptance_test
+    @original_env = ENV.to_hash
```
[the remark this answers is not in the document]
floehopper: > What did you have in mind?

Something like:

```ruby
def setup
  @original_options = ENV['MOCHA_OPTIONS']
  ENV['MOCHA_OPTIONS'] = 'verbose'
end

def teardown
  ENV['MOCHA_OPTIONS'] = @original_options
end
```

> I thought that'd be slightly more complicated and involve at least one conditional because ENV may not contain any MOCHA_OPTIONS at all originally.

If `ENV` doesn't contain any `MOCHA_OPTIONS` then `ENV['MOCHA_OPTIONS']` will return `nil`, right? So I don't see why you'd need a conditional. But maybe I'm missing something...?

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: OK. I know it's a bit pedantic, but ideally I'd like Rubocop and all the tests to pass for each commit. So as long as what you come up with achieves that, I'll be happy!

floehopper:
@nitishr I see you've done a bunch more work on this. Let me know if/when you'd like me to review anything. And in particular whether you're happy with the idea of only merging the first half of this PR as a first step...? I'm also happy to take it off your hands now you've done all the hard work if you're fed up with it!

nitishr:
OK, I think I'm now done with this PR. I've made assert_invocations less brittle. I think that assertion as well as the whole test file could be further improved, but I don't have any bright ideas atm.

As for unit tests for the new classes, I did write one for Invocation but decided to discard it because I thought it didn't do much more than to introduce an extra set of tests that're more coupled to the structure of the system than the acceptance tests. I understand that those unit tests would still be valuable to pinpoint the causes of failures more narrowly than the acceptance tests do. However, I think the risk of difficulty focusing on the cause of a failure isn't high enough to offset the cost of the unnecessary/redundant structural coupling between the implementation and unit tests.

Also, upon reflection, I'm now much more open to splitting into 2 PRs if it makes it easier for you to review or manage as I noticed there's still a number of commits for you to review. However, I'll leave it to you to decide.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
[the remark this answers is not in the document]
floehopper: I see - I hadn't considered that. In my mind the distinction isn't likely to be significant, but I'm happy to go with the way you've done it.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
```diff
@@ -35,4 +35,26 @@ def test_should_display_return_values_of_matching_invocations
...
     ], test_result.failure_message_lines
   end
+
+  def test_should_display_raised_exceptions_of_matching_invocations
+    test_result = run_as_test do
+      foo = mock('foo')
+      foo.expects(:bar).with(1).returns('a')
+      foo.stubs(:bar).with(2).returns('b').then.returns('c').then.raises(StandardError)
+
+      assert_raise(StandardError) { 3.times { foo.bar(2) } }
```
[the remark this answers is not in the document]
floehopper: Given that there's already some inconsistency in this area and I've created an issue to resolve that inconsistency, I'm happy to go with the way you've done it for now and fix it later.

lib/mocha/mock.rb
floehopper: Given that I've captured these thoughts in new issues, I'm going to mark this as resolved.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: > I respect pedantics. I'm usually a pedant myself.

😄 

> No failing commits is what I'm striving for. Sometimes it's a bit hard to tell, because although I make sure I commit only when all checks have passed locally, I don't push every commit. So, I may not realize if some intermediate commits would fail on travis for some particular setup. In most cases, I think it should be fine because when a travis build fails, I fix it using a git fixup.
> 
> Do you have a better way of ensuring this?

That's all good to know. I usually run `rake test` *locally* (which includes Rubocop checks) after every commit typically using the latest Ruby version. This runs very fast and so doesn't hinder me as much as waiting for a Travis CI build, but still gives me a lot of confidence I haven't broken anything.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: > I tried inserting this into the right earlier commit using fixup but the rebase was too painful, so I gave up. D'you have a suggestion for how I could achieve that without having to redo all subsequent commits?

Not without trying it myself. In this kind of situation I often create a new branch and cherry-pick individual commits onto it. Then when I know a given commit is going to cause conflicts I  create a *new* commit with the intention of the commit in the original branch. If the commits are all relatively small, this isn't too painful. However, I realise this can be quite time-consuming.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
nitishr: I suppose I could've been clearer. I don't commit unless rake test with my version of ruby has succeeded either. So, we have have the same process there. It's more the case of travis builds failing with other ruby versions that I wouldn't know about until I push, which I usually do only with a bunch of commits, which means that some intermediate commits could be failing for some (other) ruby version.

lib/mocha/invocation.rb
```diff
@@ -28,7 +30,7 @@ def call
 
     # @private
     def mocha_inspect
-      desc = "#{@result.mocha_inspect}"
+      desc = "\n  - #{@method_name}#{@arguments.mocha_inspect} # => #{@result.mocha_inspect}"
```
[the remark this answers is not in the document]
floehopper: Fixed in 1976f01b4d36df1e1f7b98ac637eeedd5f51e9d1 ("display arguments of matching invocations").

lib/mocha/expectation.rb
```diff
@@ -571,9 +572,13 @@ def invoke
       @yield_parameters.next_invocation.each do |yield_parameters|
         yield(*yield_parameters)
       end
-      returned_value = @return_values.next
-      @returned_values << returned_value
-      returned_value
+      begin
+        @results << @return_values.next
+      rescue Exception => e
```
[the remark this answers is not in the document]
floehopper: Fixed in 9007a97 ("record result of invocation before 'returning' it") and related commits.

lib/mocha/invocation.rb
```diff
@@ -0,0 +1,27 @@
...
+    def initialize(return_values)
+      @return_values = return_values
+    end
+
+    # @private
+    def call
+      begin
+        @result = @return_values.next
+        # rubocop:disable Lint/RescueException
+      rescue Exception => e
```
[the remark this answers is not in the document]
floehopper: Fixed in 9007a97 ("record result of invocation before 'returning' it") and related commits.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: It looks as if you've removed the changes to the test class name from 90c57c33 ("add test for yielding & returning nothing") to partially address this. 👍

floehopper:
Cool. I've created [a new version of the branch](https://github.com/freerange/mocha/compare/display-matching-invocations-alongside-expectations-v2) to which I've done the following:

* Rebased against the latest `master`
* Ran an interactive rebase executing `rake test` for each commit and fixing a small number of Rubocop violations and test failures
* Applied the remaining feedback to the relevant commits:
  * [discussion_r342006140](https://github.com/freerange/mocha/pull/394#discussion_r342006140)
  * [discussion_r342017411](https://github.com/freerange/mocha/pull/394#discussion_r342017411)
  * [discussion_r342018236](https://github.com/freerange/mocha/pull/394#discussion_r342018236)
  * [discussion_r342019257](https://github.com/freerange/mocha/pull/394#discussion_r342019257)
  * [discussion_r342020265](https://github.com/freerange/mocha/pull/394#discussion_r342020265)

I still haven't yet reviewed the later cardinality-related commits. I plan to tackle that next or at least decide whether to merge the earlier commits first.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: Cool 😎

lib/mocha/invocation.rb
```diff
@@ -0,0 +1,27 @@
+require 'mocha/raised_exception'
+
+module Mocha
+  class Invocation
+    # @private
```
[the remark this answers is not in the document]
floehopper: Resolved in https://github.com/freerange/mocha/pull/394#issuecomment-551927383.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: Resolved in https://github.com/freerange/mocha/pull/394#issuecomment-551927383.

lib/mocha/expectation.rb
floehopper: Resolved in https://github.com/freerange/mocha/pull/394#issuecomment-551927383.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: Resolved in https://github.com/freerange/mocha/pull/394#issuecomment-551927383.

test/acceptance/display_matching_invocations_alongside_expectations_test.rb
floehopper: Resolved in https://github.com/freerange/mocha/pull/394#issuecomment-551927383.

floehopper:
> I still haven't yet reviewed the later cardinality-related commits. I plan to tackle that next or at least decide whether to merge the earlier commits first.

I've had an initial look at the remaining commits. I'm not yet 100% convinced that the invocations belong "inside" the cardinality, but I need to have a bit more of a think about it.

Rather than add review comments to this PR for the later commits, I've decided to get the first commits merged and then open a new PR with the remaining commits.

floehopper:
I've now merged the first set of commits in 00f05402960995d3f8c3ac91023ed45a3e1600c6.

floehopper:
I've now opened [a 2nd PR](https://github.com/freerange/mocha/pull/410) containing the other commits. Closing this one.

Issue #173, opened 2013-12-14
Change method dispatch behaviour to that of jMock v2


floehopper:
See #171.

floehopper:
I'm no longer convinced I want to release this in v2.0, so I'm moving it out of the milestone.

PR #395, opened 2019-10-31
Change method dispatch behavior to that of jmock2

nitishr:
As suggested in #173


floehopper:
I thought I should explain that I haven't reviewed this PR yet, not because I don't welcome it, but given that it will probably require a major version bump, I've de-prioritised it in my head. I hope that's OK.

nitishr:
I understand, and am not surprised. Do you have any idea when you'd be up for a major version bump, though?

floehopper:
@nitishr I've rebased a copy of your branch against the latest `master` and pushed it up here: https://github.com/freerange/mocha/compare/change-method-dispatch-behavior-to-that-of-jmock2-without-merge-commit.

I'm definitely up for a major version release. The tricky thing about this is that I don't think it'll be possible to display deprecation warnings and so even with a major version bump, I think I might want users to be able to revert to the current behaviour by changing a configuration option. What do you think?

Ideally I'd also like to include the removal of support for some of the things I've deprecated in the current release, e.g. support for Ruby v1.8.7, and support for older versions of test-unit and minitest.

nitishr:
What if we displayed a deprecation warning for every ExpectationList that had more than one expectation? Or if any ExpectationList had more than one expectation? Haven't thought the practicality of implementation yet, but as an idea, how does that sound?

The alternative would be to provide a config switch as you've suggested. My _slight_ hesitance for that is because it'd mean more complexity not just in the implementation but also in the documentation and tests, as we'll have to implement, document and test both the behaviors. None of those is a big deal, but I think it'd be a lot simpler (and _probably_ better) to emit a deprecation warning as above.

What d'you think?

nitishr:
I've introduced a deprecation warning as I proposed above in #448.

Issue #79, opened 2012-05-01
Mocha::Expectation#in_sequence can be called without specifying any sequences


floehopper:
I can't see how it makes any sense to be able to call #in_sequence without any sequences. Doing so will impose no extra constraints on the expectations.

floehopper:
Addressed in #396. Closing.

PR #396, opened 2019-10-31
require at least one specified sequence for in_sequence

nitishr:
Addresses #79


floehopper:
I like your approach of enforcing the restriction using the method signature vs adding a conditional to raise an exception. I will get this merged shortly - thank you!

floehopper:
Merged in 9020248aae00303885dc6804be832714b7ddab47.

PR #410, opened 2019-11-08
Encapsulate invocations inside cardinality and handle Expectaction#throws scenario

floehopper:
This PR includes the 2nd set of commits from #394. The 1st set were merged in 00f05402960995d3f8c3ac91023ed45a3e1600c6.

@nitishr was the original author of all the commits in this branch. The original motivation came from #178.

### DIFF
--- a/test/acceptance/exception_rescue_test.rb
+++ b/test/acceptance/exception_rescue_test.rb
@@ -48,7 +48,7 @@ class ExceptionRescueTest < Mocha::TestCase
     assert_equal [
       'not all expectations were satisfied',
       'unsatisfied expectations:',
-      '- expected exactly once, not yet invoked: #<Mock:mock>.some_method(any_parameters)'
+      '- expected exactly once, invoked never: #<Mock:mock>.some_method(any_parameters)'
     ], test_result.failure_message_lines
   end
 end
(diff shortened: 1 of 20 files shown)


test/unit/cardinality_test.rb
```diff
@@ -1,23 +1,35 @@
 require File.expand_path('../../test_helper', __FILE__)
 require 'mocha/cardinality'
+require 'mocha/return_values'
+require 'mocha/yield_parameters'
```
floehopper: Minor: There should be a `require 'mocha/invocation'` statement here. It looks like you've added it in a later commit. I'll fix this up.

test/acceptance/exception_rescue_test.rb
```diff
@@ -48,7 +48,7 @@ def test_unsatisfied_expectation_exception_is_not_caught_by_standard_rescue
     assert_equal [
       'not all expectations were satisfied',
       'unsatisfied expectations:',
-      '- expected exactly once, not yet invoked: #<Mock:mock>.some_method(any_parameters)'
+      '- expected exactly once, invoked never: #<Mock:mock>.some_method(any_parameters)'
```
floehopper: Minor: I think "expected exactly once, invoked never" reads slightly awkwardly compared to "expected exactly once, not yet invoked" or "expected exactly once, never invoked".  However, given that it simplifies the logic, it's probably good enough.

test/unit/cardinality_test.rb
floehopper: Fixed up in 836f85b8514c841781767c1d31e923af5074d4ae.

floehopper:
Merged in 44708253962238e7ecfb14701c908f7a4a8ad6be.

PR #426, opened 2019-11-25
Document verbose option for MOCHA_OPTIONS

nitishr:


floehopper:
@nitishr Oh, sorry - I should've mentioned that I converted this to a configuration option in #421, specifically in c2b2532f36c2e9defaee71cea47cd06aa410edc6. You can see the new documentation [here](https://mocha.jamesmead.org/Mocha/Configuration.html#display_matching_invocations_on_failure=-instance_method). We could maybe use some of the copy you've written in this PR to expand the documentation for the configuration option. What do you think?

Issue #423, opened 2019-11-24
Consider moving simpler/all YARD options from .yardopts into Rake task


floehopper:
I'm not sure how we've ended up with some options defined in the `.yardopts` file and some in the Rake task. It would be good if they were all in one place if possible.

floehopper:
Closed by #429.

PR #429, opened 2019-11-26
move static yard options from rake task to .yardopts

nitishr:
This allows docs to be generated without using rake using almost the same
options as with rake (except for the dynamic options requiring ruby code - which
at this point is only the title which includes Mocha::VERSION

Addresses #423 but by taking a different (almost opposite) approach. See reasoning above.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,3 +1,4 @@
+--output-dir docs
 --template-path yard-templates
 --no-private
 lib/mocha/api.rb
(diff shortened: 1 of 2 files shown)


floehopper:
@nitishr This looks good to me.

floehopper:
Manually merged in bd9f971dfcd9a229f5b28b81f39071d4e77380dd. Closing.

PR #431, opened 2019-12-02
Reduce duplication between Mock & ObjectMethods

nitishr:


lib/mocha/mock.rb
```diff
@@ -370,5 +354,19 @@ def ensure_method_not_already_defined(method_name)
...
+      Array(method_name_or_hash).map do |*args|
+        args = args.flatten
+        method_name = args.shift
+        Mockery.instance.stub_method(object, method_name) unless object.is_a?(Mock)
+        ensure_method_not_already_defined(method_name)
+        expectation = Expectation.new(self, method_name, backtrace)
+        expectation.returns(args.shift) unless args.empty?
+        yield expectation if block
+        @expectations.add(expectation)
+      end.last
```
floehopper: Although I appreciate that this reduces the duplication in the `Mock` class, I'm not convinced that the resultant code is any easier to follow. In fact I think, if anything, it's a bit harder to follow! Let me have a think about how we might better remove the duplication.

floehopper:
I've added a comment inline relating to the most significant change in this PR. While I agree that removing the duplication is good, I wonder whether there are some better abstractions we could use so the resultant code is easier to understand.

lib/mocha/mock.rb
nitishr: Thanks for your review, @floehopper. Would you mind elaborating what aspects of this change you aren't convinced about and/or which ones you think make it harder to follow? If I understand your concerns more specifically, we might be able to come up with a better solution together.

To start with, I can share a bit more of my thinking here...

If it's the `anticipates` abstraction, I'll admit that I was initially unsure of it as well (and still am, but a lot less than I initially was). The reason I was unsure was I wondered if the need for me to look up (and cite the dictionary definitions in the [PR description](https://github.com/freerange/mocha/pull/431#issue-347805935)) was a sign of the concept being too smart/cute/nuanced.

So, I let it mull for a few days in my head and reached the conclusion that `anticipates` is probably a better abstraction covering both an expectation and an allowance (stub, in mocha's terminology). I've always found expectation to be too strong a term to represent an allowance. (Anticipation might be too strong a term, too - but it does seem _less_ strong than expectation.) So, in the long term, I would even consider/suggest considering renaming, for instance, the class `Expectation` to `Anticipation` (and maybe introduce subclasses - `Expectation` and `Allowance`). I have of course not fully analyzed that chain of reasoning, and certainly not considered it from an API compatibility point of view. This was only trying to clarify the concepts and their nuances for myself.

Having said that, I'm not a native English speaker and may well have misinterpreted the shades of meanings of the various terms. So, I'd appreciate any corrections to my understanding or thinking.

floehopper: Sorry for the slow reply. I'll try to explain my reservations:

* Although the `Mock#anticipates` method has removed all the duplication between `ObjectMethods#expects`, `ObjectMethods#stubs`, `Mock#expects` & `Mock#stubs` (which is great BTW!), it's now really quite a long and "dense" method  and, what's more, three of the lines are only executed conditionally. Coming at the final version of this method definition cold, I found it quite hard to follow.

* I'm not completely convinced that the inlining of the `ArgumentIterator` is a net gain. Encapsulating the common processing of the arguments for the four API methods in a single place still makes sense to me, even though its exact abstraction may not be quite right. And having to work out what the combination of calls to `Kernel#Array`, `Array#map`, `Array#flatten`, `Array#shift` & `Array#last` is doing adds quite a lot of cognitive overhead to understanding the method definition.

* The naming of `Mock#anticipates` doesn't bother me too much, although I don't think it is _quite_ right. I wonder whether it would make more sense if we made an `Expectation` have a cardinality of `Cardinality.at_least(0)` (i.e. stubbing behaviour) by default, inline `Mock#anticipates` into `Mock#stubs`, and call `Mock#stubs` from `Mock#expects` passing a block setting the expectation cardinality to `Cardinality.exactly(1)` (i.e. default expecting behaviour). It feels as if the expecting behaviour is a superset of the stubbing behaviour. What do you think...?

Does make sense? I'm sorry I've struggled a bit to explain my thinking!

I've had time to go through the commits one-by-one in more detail this afternoon and I'd like to mention that I found them very easy to follow - so thank you for taking so much care to curate the commits and write useful commit messages.

Having looked through the commits in more detail, I'm inclined to have a go at merging a version of this PR, albeit with some changes as per my comments above. Is that OK with you?

nitishr: That's very helpful, @floehopper. I now understand your concerns much better.

I like the suggestion of `expects` constraining the `Expectation`, rather than `stubs` loosening it. Part of the reason I extracted `anticipates` rather than reuse `expects`, BTW, was to avoid changing the public API of the existing methods, even though the changes would be fully backward compatible (i.e. accepting but not expecting a block or optional arguments). So, that's a consideration.

I'd be happy with you merging a modified version of the PR with the changes you feel comfortable with. We could iterate on top of that for the more controversial/difficult aspects.

floehopper: 👍

nitishr: @floehopper, I'm planning to push another commit which I think will alleviate some of your concerns, while at the same time, improving mocha's behavior with expectation chaining by making it more consistent across single-method and multiple-method stubs/expects.

The code in that commit or even the approach could possibly be improved, but I'd like to convey the general idea/direction before I go too far.

Is that OK? If you'd rather that I wait for you to finish merging (a version of) this PR, that's fine, too. Just let me know.

floehopper: > BTW, was to avoid changing the public API of the existing methods, even though the changes would be fully backward compatible (i.e. accepting but not expecting a block or optional arguments). So, that's a consideration.

Good point! 😄 

> Is that OK? If you'd rather that I wait for you to finish merging (a version of) this PR, that's fine, too. Just let me know.

Yes, that's fine - please go ahead!

nitishr: Done.

I'm completely fine taking any or all of the newly added commits to a different PR so they can be discussed and reviewed on their own (or altogether dropping any/all of them). The newly introduced `ExpectationSetting` may need some redesign, but I hope the changes convey the general concept and approach.

nitishr: > I like the suggestion of expects constraining the Expectation, rather than stubs loosening it.

Looking at the code after my last few updates, I think the current form expresses the intent better than the change we were thinking of.

```ruby
def stubs(stubbed_methods_vs_return_values)
  expects(stubbed_methods_vs_return_values).at_least(0)
end
```

seem more intention-revealing and logical than

```ruby
def expects(stubbed_methods_vs_return_values)
  stubs(stubbed_methods_vs_return_values).exactly(1)
end
```

nitishr:
Please merge this to stop me pushing more on this branch :-)

lib/mocha/composite_expectation.rb
```diff
@@ -0,0 +1,18 @@
...
+  class CompositeExpectation
+    attr_reader :expectations
+
+    def initialize(expectations)
+      @expectations = expectations
+    end
+
+    %w[
+      times twice once never at_least at_least_once at_most at_most_once
+      with yields multiple_yields returns raises throws then when in_sequence
```
nitishr: Remember to add with_block_given and with_no_block_given when merging

lib/mocha/mock.rb
```diff
@@ -109,14 +109,15 @@ class Mock
     #   object.expects(:expected_method_one).returns(:result_one)
     #   object.expects(:expected_method_two).returns(:result_two)
     def expects(method_name_or_hash, backtrace = nil)
-      iterator = ArgumentIterator.new(method_name_or_hash)
-      iterator.each do |*args|
+      CompositeExpectation.new(Array(method_name_or_hash).map do |*args|
```
nitishr: Looking at the documentation:

`@return [Expectation] last-built expectation which can be further modified by methods on {Expectation}`

This introduces a breaking change. So, even if we agree that the new behavior is preferable, we'd have to think about deprecating the old behavior.

There are no existing tests to check for the documented behavior, though.

lib/mocha/mock.rb
floehopper: > @floehopper, I'm planning to push another commit which I think will alleviate some of your concerns, while at the same time, improving mocha's behavior with expectation chaining by making it more consistent across single-method and multiple-method stubs/expects.
> 
> The code in that commit or even the approach could possibly be improved, but I'd like to convey the general idea/direction before I go too far.
> 
> Is that OK? If you'd rather that I wait for you to finish merging (a version of) this PR, that's fine, too. Just let me know.

Sorry for the slow reply. Thanks for adding the extra commit - I'm going to take a look at it shortly.

PR #434, opened 2019-12-04
Reintroduce allow check removed in 948a0c2

### LINKED ISSUE #432: Stubbing methods on classes that dynamically define methods don't work anymore
If the `respond_to?` of an object depends on its internal state and we use `any_instance` it is not possible to stub any method of this object.

```irb
>> require 'mocha/api'
=> true
>> class Foo
>>   def initialize(attrs = {})
>>     @attributes = attrs
>>     end
>>   def respond_to?(method, *args)
>>     @attributes.key?(method) ? @attributes[method] : super
>>     end
>>   end
=> :respond_to?
>> include Mocha::API
=> Object
>> Foo.any_instance.stubs(:foo)
Traceback (most recent call last):
       12: from /opt/rubies/2.6.5/bin/irb:23:in `<main>'
       11: from /opt/rubies/2.6.5/bin/irb:23:in `load'
       10: from /opt/rubies/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        9: from (irb):13
        8: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/object_methods.rb:127:in `stubs'
        7: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/argument_iterator.rb:13:in `each'
        6: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/object_methods.rb:129:in `block in stubs'
        5: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/mockery.rb:131:in `on_stubbing'
        4: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/mockery.rb:154:in `check'
        3: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/mockery.rb:132:in `block in on_stubbing'
        2: from /Users/rafaelfranca/.gem/ruby/2.6.5/gems/mocha-1.10.0/lib/mocha/class_methods.rb:30:in `respond_to?'
        1: from (irb):7:in `respond_to?'
NoMethodError (undefined method `key?' for nil:NilClass)
```

This happen because we are trying to `allocate` the object what leaves the object with the wrong internal state. It is common to people to setup all internal state of an object using the `initialize` method, so I don't think anyone setup internal state of objects defining a `allocate`.

nitishr:
Fixes #432 by short-circuiting the method_exists? check in case the treatment of
the action is to allow it.

### DIFF
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -151,14 +151,11 @@ module Mocha
     private
 
     def check(action, description, method_signature, backtrace = caller)
-      return if block_given? && !yield
+      treatment = Mocha.configuration.send(action)
+      return if (treatment == :allow) || (block_given? && !yield)
       message = "stubbing #{description}: #{method_signature}"
-      case Mocha.configuration.send(action)
-      when :warn
-        logger.warn(message)
-      when :prevent
-        raise StubbingError.new(message, backtrace)
-      end
+      raise StubbingError.new(message, backtrace) if treatment == :prevent
+      logger.warn(message) if treatment == :warn
     end
 
     def expectations
--- a/test/acceptance/stubbing_non_existent_any_instance_method_test.rb
+++ b/test/acceptance/stubbing_non_existent_any_instance_method_test.rb
@@ -76,6 +76,22 @@ class StubbingNonExistentAnyInstanceMethodTest < Mocha::TestCase
     assert_passed(test_result)
   end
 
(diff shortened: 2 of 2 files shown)


floehopper:
This looks good to me. Thanks for fixing the problem!

Issue #436, opened 2019-12-10
`.yields()` failing when the caller does not provide a block breaks legitimate tests


yemartin:
## `.yields()` failing when the caller does not provide a block breaks legitimate tests

I have a library with an expensive method (API calls) that both yields and returns. In some tests, both the yielding and the returning aspects are used, so I need to stub both. This used to work fine in 1.9.0, but broke when upgrading to 1.10.1. I traced it down to this in the 1.10.0.alpha release notes (emphasis mine):

> Make Expectation#yields & Expectation#multiple_yields fail when the caller of the stubbed method does not provide a block. This is a change to an undocumented aspect of the public API's behaviour. **If this causes your tests to fail, then fix it by removing the unnecessary call to Expectation#yields or Expectation#multiple_yields** - thanks to @nitishr (#382)

But my calls to `.yields` are legitimate and necessary, I cannot just remove them. So I am left with broken tests, and I am not sure how to fix them. Maybe the change in #382 needs to be reevaluated?

## How to reproduce

Here is a code sample that illustrates how one may need both `.returns` and `.yields` in the same test:

```ruby
# my_service.rb

class MyService
  def self.expensive_method
    # Some very expensive calls here
    yield "yielded" if block_given?
    return "returned"
  end

  def self.call
    results = []
    expensive_method do |yielded|
      results << yielded
    end
    results << expensive_method
  end
end
```

```ruby
# test_my_service.rb

class TestMyService < Minitest::Test

  def test_returns
    MyService.stubs(:expensive_method).returns("returned")
    assert_equal "returned", MyService.expensive_method
  end

  def test_yields
    MyService.stubs(:expensive_method).yields("yielded")
    MyService.expensive_method do |yielded|
      assert_equal "yielded", yielded
    end
  end


  def test_both_returns_and_yields
    MyService
      .stubs(:expensive_method)
      .returns("returned")
      .yields("yielded")
    call_result = MyService.call
    assert_includes call_result, "returned"
    assert_includes call_result, "yielded"
  end
end
```

### With mocha 1.9.0:

```
# Running:

...

Finished in 0.001714s, 1750.2917 runs/s, 3500.5834 assertions/s.

3 runs, 6 assertions, 0 failures, 0 errors, 0 skips
```

### With mocha 1.10.1:

```
# Running:

E..

Finished in 0.001527s, 1964.6365 runs/s, 1309.7577 assertions/s.

  1) Error:
TestMyService#test_both_returns_and_yields:
[...]

PR #437, opened 2019-12-10
Optionally reinstate v1.9 behaviour of Expectation#yields & Expectation#multiple_yields when no block given

floehopper:
This is in relation to #436.

In v1.9 if `Expectation#yields` or `Expectation#multiple_yields` was called on an expectation, but no block was given when the method was invoked, the instruction to yield was ignored. Since v1.10, a `LocalJumpError` is raised.

The old undocumented behaviour is now available via the new `retain_undocumented_behaviour_from_v1_9` configuration option, albeit with a deprecation warning. So you'll need to do something like the following to make use of it:

```ruby
Mocha.configure do |c|
  c.retain_undocumented_behaviour_from_v1_9 = true
end
```

### DIFF
--- a/lib/mocha/configuration.rb
+++ b/lib/mocha/configuration.rb
@@ -250,6 +250,9 @@ module Mocha
 
     # Retain undocumented behaviour from v1.9
     #
+    # Previously if {Expectation#yields} or {Expectation#multiple_yields} was called on an expectation, but no block was given when the method was invoked, the instruction to yield was ignored.
+    # Now a LocalJumpError is raised.
+    #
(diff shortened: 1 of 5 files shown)


yemartin:
Thanks a lot for the patch @floehopper! We have tested it and it does work as expected, restoring the old behavior and allowing our tests to pass again 👍

floehopper:
@yemartin 

> Thanks a lot for the patch @floehopper! We have tested it and it does work as expected, restoring the old behavior and allowing our tests to pass again 👍

Great - thanks for letting me know.

floehopper:
Closing in favour of #438.

PR #438, opened 2019-12-11
Optionally reinstate undocumented behaviour from v1.9

floehopper:
This introduces a new configuration option (`reinstate_undocumented_behaviour_from_v1_9`) to reinstate a couple of bits of undocumented behaviour from v1.9 which were changed in v1.10 without any prior deprecation warning. It turns out that a number of projects were depending on this behaviour and I'm keen to provide them with a way to upgrade to the latest version of Mocha, but see deprecation warnings for the behaviour which has changed.

* The behaviour of `API#mock`, `API#stub` and `API#stub_everything` when called with a symbol as the first argument.

* The behaviour of `Expectation#yields` and `Expectation#multiple_yields` when the stubbed method is called without a block.

This supersedes #437, because it includes the same functionality albeit with a different configuration option name. It should therefore address #436.

It also includes work I did in [this branch][1] in response to a private email regarding the behaviour of `API#mock`, `API#stub` and `API#stub_everything`.

[1]: https://github.com/freerange/mocha/compare/reinstate-undocumented-behaviour-changed-in-v1.10

### DIFF
--- a/lib/mocha/api.rb
+++ b/lib/mocha/api.rb
@@ -70,15 +70,15 @@ module Mocha
       if Mocha.configuration.reinstate_undocumented_behaviour_from_v1_9?
         if arguments.first.is_a?(Symbol)
           method_name = arguments[0]
-          Deprecation.warning([
+          Deprecation.warning(
             "Explicitly include `#{method_name}` in Hash of expected methods vs return values,",
-            "e.g. `mock(:#{method_name} => nil)`."
-          ].join(' '))
+            " e.g. `mock(:#{method_name} => nil)`."
+          )
           if arguments[1]
-            Deprecation.warning([
+            Deprecation.warning(
               "In this case the 2nd argument for `mock(:##{method_name}, ...)` is ignored,",
-              'but in the future a Hash of expected methods vs return values will be respected.'
-            ].join(' '))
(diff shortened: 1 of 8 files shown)


floehopper:
@nitishr Does this sound OK to you? I realise it feels like a bit of a backward step (and adds some duplication & complexity), but hopefully much of the code for the now deprecated behaviour can be safely removed in a new release.

nitishr:
LGTM, except for the duplication introduced in API#{mock,stub,stub_everything} in 2f79b58. Could we try and reduce that? I can give it a go if you'd like.

👍 for storing the block on invocation in 72dac85.

floehopper:
> LGTM, except for the duplication introduced in API#{mock,stub,stub_everything} in [2f79b58](https://github.com/freerange/mocha/commit/2f79b58fed8f360416c00ad6748ed405f6054267). Could we try and reduce that? I can give it a go if you'd like.

To be honest, I'm not too bothered about the duplication, because I plan to remove it soon. There are quite a few small differences which I think might make it awkward to remove the duplication, so I don't think it's worth the time.

> 👍 for storing the block on invocation in [72dac85](https://github.com/freerange/mocha/commit/72dac85f0aced2c9a27a3de3d6db3e0f46e80668).

Cool! 👍

PR #439, opened 2019-12-11
delegate yields to multiple_yields

nitishr:
and delete XxYields classes


floehopper:
@nitishr:

I think this is a useful simplification. Sometimes I'm a fan of more classes than strictly necessary if it makes the implementation clearer, but I think in this case you haven't really lost any clarity by removing them.

It might make sense to rename `YieldParameters#multiple_add` to just `YieldParameters#add` as part of this PR. What do you think?

nitishr:
Good suggestion. Done.

I tend to use the [four elements of simple design](https://blog.jbrains.ca/permalink/the-four-elements-of-simple-design) as a heuristic:

- Passes its tests
- Minimizes duplication
- Maximizes clarity
- Has fewer elements

floehopper:
Thanks. I'll get this merged.

floehopper:
I hope you don't mind, but I've re-written a bunch of the commit messages slightly, partly as an exercise for me to understand what each commit was doing, and merged the resulting branch manually in [this commit](https://github.com/freerange/mocha/commit/7f2767f5b02e22e457348345a2063948c8930f56). Closing.

PR #441, opened 2019-12-13
Add Expectation#with_block_given & Expectation#with_no_block_given

floehopper:
This relates to #436 & #382.

Since #382, a `LocalJumpError` is raised if an invocation does not supply a block, but the matching expectation includes an instruction to yield. Until now there has been no way to constrain an expectation to match based on whether the invocation supplies a block or not. This PR makes this possible.

These changes have allowed me to improve the deprecation warning in `Invocation#call` to give a concrete suggestion on how to avoid the deprecated functionality.

@nitishr I've included a change `Invocation#call_description` to append " { ... }" to give an indication that the invocation supplied a block. I'm not 100% sure this is the best string representation - so I'm open to ideas on this front. I'm also not sure whether I should add something similar to `Invocation#short_call_description` too...?

### DIFF
--- a/lib/mocha/block_matcher.rb
+++ b/lib/mocha/block_matcher.rb
@@ -1,0 +1,31 @@
+module Mocha
+  module BlockMatchers
+    class OptionalBlock
+      def match?(_actual_block)
+        true
+      end
+
+      def mocha_inspect; end
+    end
+
+    class BlockGiven
+      def match?(actual_block)
+        !actual_block.nil?
+      end
+
+      def mocha_inspect
+        'with block given'
+      end
+    end
+
+    class NoBlockGiven
+      def match?(actual_block)
+        actual_block.nil?
+      end
+
+      def mocha_inspect
+        'with no block given'
+      end
+    end
+  end
+end
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -10,6 +10,7 @@ require 'mocha/in_state_ordering_constraint'
 require 'mocha/change_state_side_effect'
 require 'mocha/cardinality'
 require 'mocha/configuration'
+require 'mocha/block_matcher'
 
 module Mocha
   # Methods on expectations returned from {Mock#expects}, {Mock#stubs}, {ObjectMethods#expects} and {ObjectMethods#stubs}.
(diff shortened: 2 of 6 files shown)


lib/mocha/expectation.rb
```diff
@@ -224,6 +225,44 @@ def with(*expected_parameters, &matching_block)
...
+    #   object = mock()
+    #   object.expects(:expected_method).with_block_given
+    #   object.expected_method { 1 + 1 }
+    #   # => verify succeeds
+    #
+    #   object = mock()
+    #   object.expects(:expected_method).with_block_given
+    #   object.expected_method
+    #   # => verify fails
+    def with_block_given
```
nitishr: Would `with_a_block` and `without_a_block` be better names?

lib/mocha/invocation.rb
```diff
@@ -53,7 +55,9 @@ def arguments
     end
 
     def call_description
-      "#{@mock.mocha_inspect}.#{@method_name}#{ParametersMatcher.new(@arguments).mocha_inspect}"
+      description = "#{@mock.mocha_inspect}.#{@method_name}#{ParametersMatcher.new(@arguments).mocha_inspect}"
+      description << ' { ... }' unless @block.nil?
```
nitishr: We could make this more informative as below:

1. { |arg1, arg2| <block body> } relying on Proc#arity
2. { |x, y, *other| <block body> } relying on Proc#parameters

2 would be smarter but more complex. 1 might be a good enough solution. I expect call_description to be a rarely used feature anyway. So, I'd avoid being smart at the cost of complexity until a need is validated.

Even the current form might be good enough.

In any case, I suggest replacing ... with something like <block body>, making it clearer that we're representing a block.

nitishr:
Looks great. I made a few small suggestions inline, and have a question for consideration: should we raise an error when `.yields` is called on an expectation `.with_no_block_given`? Probably not, since one might be intentionally trying to simulate a situation that'll result in a LocalJumpError (although I can't think of why).

lib/mocha/block_matcher.rb
```diff
@@ -0,0 +1,31 @@
+module Mocha
+  module BlockMatchers
+    class AnyBlock
```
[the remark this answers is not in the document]
floehopper: Agreed. I went backwards and forwards on the naming, but you're right that `OptionalBlock` is less ambiguous. 👍

lib/mocha/invocation.rb
floehopper: Hmm. Thanks. I did consider doing something with `Proc#arity`. However, I think what I have is good enough for now and I don't think anyone should start relying on the exact format, so we should be free to change it at a later date.

I hadn't come across `Proc#parameters`. That's a nice addition to the language, but I assume it's only been introduced relatively recently...?

lib/mocha/expectation.rb
floehopper: I went backward and forwards on the naming of these methods. I quite like the fact that the names I've chosen mirror the naming of the `Kernel#block_given?` method. However, I'll give your suggestion some thought.

floehopper:
> Looks great. I made a few small suggestions inline, and have a question for consideration: should we raise an error when `.yields` is called on an expectation `.with_no_block_given`? Probably not, since one might be intentionally trying to simulate a situation that'll result in a LocalJumpError (although I can't think of why).

Hmm. That's an interesting thought. Although on balance, I think I'm inclined to agree with you (i.e. "probably not").

floehopper:
Rebasing against latest `master` and force-pushing.

lib/mocha/block_matcher.rb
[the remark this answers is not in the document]
floehopper: I've now made this change locally and I'll push up the re-written commits when I've addressed the other feedback.

lib/mocha/expectation.rb
floehopper: I've talked to a couple of other people to ask their opinion on this naming and I've decided to stick with the current naming, i.e. `with_block_given` & `with_no_block_given`.

lib/mocha/invocation.rb
floehopper: I'm going to stick with what I've got for now, i.e. "{ ... }".

Issue #443, opened 2019-12-16
NoMethodError: undefined method `map'


floehopper:
I noticed [this PR][1] in which they [tried to use v1.10.2][2] to fix the issue in #436, but ran into [another problem][3].

```
NoMethodError: undefined method `map' for #<String:0x000055f157c8de00>
    lib/mocha/parameters_matcher.rb:31:in `matchers'
    lib/mocha/parameters_matcher.rb:24:in `mocha_inspect'
    lib/mocha/invocation.rb:29:in `block in call'
    lib/mocha/multiple_yields.rb:11:in `block in each'
    lib/mocha/multiple_yields.rb:10:in `each'
    lib/mocha/multiple_yields.rb:10:in `each'
    lib/mocha/invocation.rb:23:in `call'
    lib/mocha/expectation.rb:571:in `invoke'
    lib/mocha/mock.rb:315:in `method_missing'
```

[1]: https://github.com/apiology/quality/issues/121
[2]: https://github.com/apiology/quality/commit/ec9b4ce0f5bb7f0db84316c6bb4a4d57bb924f4f
[3]: https://circleci.com/gh/apiology/quality/351

floehopper:
I'm currently investigating.

floehopper:
See possible fix in #444

floehopper:
Fixed in #444.

PR #444, opened 2019-12-16
Fix issue with non-Array arguments passed to Expectation#multiple_yields

floehopper:
This was highlighted when investigating [this issue][1]. It turned out that this project was calling `Expectation#multiple_yields` with non-Array arguments, mostly/all Strings. See #443 for details.

While the documentation said that each argument should be an `Array`, it turned out that in most cases the functionality still worked with non-`Array` arguments, but I suspect this was by accident rather than
design. One exception to this was if an argument was a `Hash`, then it  ended up being passed to the block as an `Array` of two-element Arrays.

[This code][2] which is used to optionally displays invocations was assuming that `yield_args` was always an `Array`, but this was not true. Similarly [this code][3] which is used to display a deprecation warning was also assuming that `yield_args` was always an `Array`. And it was the latter which was causing [the `NoMethodError` in the apiology/quality project][4].

I've hopefully fixed the underlying problem by wrapping any non-`Array` argument in an `Array`. Note that using `Kernel#Array` was not an option,  because this didn't give the correct behaviour for `Hash` arguments.

There is an argument to say that this was undocumented behaviour, but I think that seems unnecessarily strict in this case.

[1]: https://github.com/apiology/quality/issues/121
[2]: https://github.com/freerange/mocha/blob/1070fc02015a8ea36ad50782f80b557813595c02/lib/mocha/invocation.rb#L24
[3]: https://github.com/freerange/mocha/blob/1070fc02015a8ea36ad50782f80b557813595c02/lib/mocha/invocation.rb#L29
[4]: https://circleci.com/gh/apiology/quality/351

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -261,7 +261,7 @@ module Mocha
 
     # Modifies expectation so that when the expected method is called, it yields multiple times per invocation with the specified +parameter_groups+ (even if no block is provided, in which case yielding will result in a +LocalJumpError+).
     #
-    # @param [*Array<Array>] parameter_groups each element of +parameter_groups+ should iself be an +Array+ representing the parameters to be passed to the block for a single yield.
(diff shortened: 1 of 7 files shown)


floehopper:
Manually merged in 215f8e1301fb8ea7544f0937a0cc7878e033a621. Closing.

PR #451, opened 2019-12-26
replace original_{method,visibility} attrs w/ ivars

nitishr:
...since the attrs were only used internally and were giving rise to a
'warning: private attribute?'


floehopper:
Looks good to me. 👍

PR #452, opened 2019-12-27
set up {Moch,Stubb}aExampleTest as acceptance tests

nitishr:
These were failing when run standalone


floehopper:
Looks good to me. 👍

PR #454, opened 2020-01-03
there is no unsatisfied expectation exception

nitishr:
... and nowhere to catch it

### DIFF
--- a/test/acceptance/exception_rescue_test.rb
+++ b/test/acceptance/exception_rescue_test.rb
@@ -37,17 +37,4 @@ class ExceptionRescueTest < Mocha::TestCase
     assert_failed(test_result)
     assert_equal 'unexpected invocation: #<Mock:mock>.some_method()', test_result.failure_message_lines[0]
   end
-
-  def test_unsatisfied_expectation_exception_is_not_caught_by_standard_rescue
-    test_result = run_as_test do
-      mock = mock('mock')
-      mock.expects(:some_method)
-    end
-    assert_failed(test_result)
-    assert_equal [
-      'not all expectations were satisfied',
-      'unsatisfied expectations:',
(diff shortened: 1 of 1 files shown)


floehopper:
Although you are correct that there is no exception called anything along the lines of "unsatisfied expectation exception", I believe this was intended to refer to [the exception raised in `Mockery#verify`](https://github.com/freerange/mocha/blob/8751dcbda08650a339932d86c9f24d076c167300/lib/mocha/mockery.rb#L93), i.e. [`ExpectationError`](https://github.com/freerange/mocha/blob/8751dcbda08650a339932d86c9f24d076c167300/lib/mocha/expectation_error.rb#L7) when using Test::Unit and [`MiniTest::Assertion`](https://github.com/freerange/mocha/blob/8751dcbda08650a339932d86c9f24d076c167300/lib/mocha/integration/mini_test/adapter.rb#L26) when using MiniTest. 

When an unexpected invocation is received by a mock object, it raises the same type of exception (see [`Mock#raise_unexpected_invocation_error`](https://github.com/freerange/mocha/blob/8751dcbda08650a339932d86c9f24d076c167300/lib/mocha/mock.rb#L380)). Thus the scenario for an unexpected invocation earlier in the same test case gives us some related coverage.

I believe the reason why we don't want the exception class to be rescued by a standard `rescue` is that the code under test could legitimately include such a rescue block. If the Mocha-generated exception was rescued by the code under test, the test failure might be swallowed or a confusing error message might be displayed. It's much less legitimate for code under test to rescue `Exception` and so it's preferable if the Mocha-generated exception is a direct subclass of `Exception` and not `StandardError`.

Having said all the above, I think you've correctly identified that the test in question is not really testing anything useful that isn't tested elsewhere, so I'm happy to accept this PR. Although I might add some of my notes from this comment to the commit note before merging.

PR #462, opened 2020-01-16
method_missing won't throw if ExpectationList empty

### LINKED ISSUE #461: Race condition when using `stubs` in JRuby
A common example of using _mocha_ is to stub Time to return a fixed time.
For instance,
```ruby
now = Time.now
Time.stubs(:now).returns(now)
```

However if there is a separate thread running which tries to invoke the stubbed method (specifically on non-GIL interpreters like JRuby), we've encountered a race condition in that the mocha will raise an exception that the expectations have not been met.

here is a simple example:

```ruby
  def test_mocha_thread_safety
    Thread.new do
      500.times do
        # just loop and call Time#now a bunch while the thread above keeps stubbing
        Time.now
      end
    end

    now = Time.now
    500.times do
      # the two mechanisms involve different stubbing code, wso lets test both
      Time.stubs(:now).returns(now)
    end

  end
```

Looking through the code; the result seems to be due to the `mock.rb` not being protected with a critical section.

```ruby
    def stubs(method_name_or_hash, backtrace = nil)
      iterator = ArgumentIterator.new(method_name_or_hash)
      iterator.each do |*args|
        method_name = args.shift
        ensure_method_not_already_defined(method_name)
        expectation = Expectation.new(self, method_name, backtrace)
        expectation.at_least(0)
        expectation.returns(args.shift) unless args.empty?
        @expectations.add(expectation)
      end
    end
```

Specifically if `@expectations.add(expectation)` is not executed but the method has been _stubbed_, then the `method_missing` in _mock_ will raise an exception since there is no expectation to match against.

stubs + method_missing need to share the same mutex and become a critical section
or
method_missing should not raise if expectations is empty

fzakaria:
This is my attempt to fix https://github.com/freerange/mocha/issues/461
Although _mocha_ does not claim to be thread-safe; this change slowly
makes it more resilient to race conditions through use of mocha.

A method on an object/class may be stubbed & invoked prior to the
expectation being added to the list. In such a scenario, the
method_missing in `mock.rb` would raise an exception for unknown
invocation.

Smply add a if-guard that checks that the expectation list is not empty.

fixes #461

### DIFF
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -312,7 +312,7 @@ module Mocha
       invocation = Invocation.new(self, symbol, *arguments, &block)
       if (matching_expectation_allowing_invocation = all_expectations.match_allowing_invocation(invocation))
         matching_expectation_allowing_invocation.invoke(invocation)
-      elsif (matching_expectation = all_expectations.match(invocation)) || (!matching_expectation && !@everything_stubbed)
+      elsif all_expectations.any? && (matching_expectation = all_expectations.match(invocation)) || (!matching_expectation && !@everything_stubbed)
         raise_unexpected_invocation_error(invocation, matching_expectation)
       end
     end


floehopper:
@fzakaria Thanks for taking the time to open this PR. Given that an instance of `Mock` has a single instance of `ExpectationList` shared across all method calls, I'm not sure this is a generic solution to your problem, is it? i.e. if another method on the same `Mock` instance was stubbed prior to your scenario, the instance of `ExpectationList` would not be empty.

fzakaria:
That makes sense.
I added a more thorough response in the issue as well -- thank you.

floehopper:
👍

PR #463, opened 2020-01-17
stubbee -> stubba_object, mock_owner -> stubbee

nitishr:
Rename: stubbee -> stubba_object, mock_owner -> stubbee

stubbee is the thing that's stubbed - more specifically, the object on
which .stubs or .mocks is called. (Therefore, it's also the object that
holds a reference to mocha. That's why it was called mock_owner earlier.
But that's an implementation detail that doesn't have to reflect in the
naming)

stubba_object is the object through which we get hold of the
stubba_class on which stubbed methods reside. In case of AnyInstance,
that class is the stubba_object itself, while in case of Instance, it's
the singleton class of the stubba_object.

This is consistent with the meaning and usage of stubba_object and
stubba_class methods in ObjectMethods and AnyInstance.

### DIFF
--- a/lib/mocha/any_instance_method.rb
+++ b/lib/mocha/any_instance_method.rb
@@ -6,16 +6,16 @@ module Mocha
   class AnyInstanceMethod < StubbedMethod
     private
 
-    def mock_owner
-      stubbee.any_instance
+    def stubbee
+      stubba_object.any_instance
     end
 
     def stubbee_method(method_name)
-      stubbee.instance_method(method_name)
+      stubba_object.instance_method(method_name)
     end
 
     def original_method_owner
-      stubbee
+      stubba_object
     end
   end
 end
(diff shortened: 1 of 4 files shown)


floehopper:
@nitishr Sorry for the slow response - I've been busy with client work. Thanks for all your ongoing work on Mocha. I'll take a look at all your open PRs as soon as I can.

nitishr:
Sure, no rush. If I can do anything to make it easier, please let me know.

nitishr:
@floehopper I've resurrected this tiny PR that makes a small refinement to the 'ubiquitous language'.

floehopper (approved):
This all looks good to me. Thanks for reviving it! 👍

floehopper:
I've backed out an unrelated formatting change in #751.

floehopper:
Superseded by #751. Closing.

PR #469, opened 2020-02-17
removing the Spanish Inquisition special case :(

### LINKED ISSUE #468: Consider removing the Spanish Inquisition special case
https://en.wikipedia.org/wiki/The_Spanish_Inquisition_(Monty_Python)

I think it's time may have passed...

nitishr:
Closes #468


nitishr:
This has now been incorporated in #431.

PR #474, opened 2020-03-06
Remember invocations

### LINKED ISSUE #473: Change in partial expectation evaluation time?
It appears that in 1.10.0, the time at which partial expectations start recording changed, but I don't see anything in the release notes that explains what I'm seeing. I suspect that <1.10.0, they started recording at the time of partial definition, but >=1.10.0 they start recording only when they've been finalized with a quantifier.  For example, see this simplified snippet of some code found in the wild:

```
before do
  exp = Foo.any_instance.expects(:action)
end

it 'invokes' do
  Foo.action
  exp.twice
  Foo.action
end
```

Moving the `twice` up a line works regardless of version.

Is this expected? Is there another explanation?

nitishr:
Fixes #473. Earlier, whenever you qualified an Expectation with a cardinality, a new Cardinality instance would be replace the old one. 59454a8 moved the invocations recording to Cardinality, which then meant that the invocations recorded prior to updating cardinality would get discarded along with the Cardinality instance, effectively being forgotten.
    
We now update the quantifiers of the same Cardinality instance for each Expectation in order to persist the recording of invocations across updates to cardinality.

### DIFF
--- a/lib/mocha/cardinality.rb
+++ b/lib/mocha/cardinality.rb
@@ -2,31 +2,28 @@ module Mocha
   class Cardinality
     INFINITY = 1 / 0.0
 
-    class << self
-      def exactly(count)
-        new(count, count)
-      end
-
-      def at_least(count)
-        new(count, INFINITY)
-      end
-
-      def at_most(count)
-        new(0, count)
-      end
-
-      def times(range_or_count)
-        case range_or_count
-        when Range then new(range_or_count.first, range_or_count.last)
-        else new(range_or_count, range_or_count)
-        end
-      end
+    def initialize(required = 0, maximum = INFINITY)
+      update(required, maximum)
+      @invocations = []
     end
 
-    def initialize(required, maximum)
-      @required = required
-      @maximum = maximum
-      @invocations = []
+    def exactly(count)
+      update(count, count)
+    end
+
+    def at_least(count)
+      update(count, INFINITY)
+    end
+
+    def at_most(count)
+      update(0, count)
+    end
+
+    def times(range_or_count)
+      case range_or_count
+      when Range then update(range_or_count.first, range_or_count.last)
+      else update(range_or_count, range_or_count)
+      end
     end
 
     def <<(invocation)
@@ -94,6 +91,12 @@ module Mocha
       when 2 then 'twice'
       else "#{number} times"
       end
+    end
+
+    def update(required, maximum)
+      @required = required
+      @maximum = maximum
+      self
     end
 
     def infinite?(number)
--- a/lib/mocha/cardinality.rb
+++ b/lib/mocha/cardinality.rb
@@ -62,21 +62,21 @@ module Mocha
       if allowed_any_number_of_times?
         'allowed any number of times'
       elsif required.zero? && maximum.zero?
-        "expected #{times(maximum)}"
+        "expected #{count(maximum)}"
       elsif required == maximum
-        "expected exactly #{times(required)}"
+        "expected exactly #{count(required)}"
       elsif infinite?(maximum)
-        "expected at least #{times(required)}"
+        "expected at least #{count(required)}"
       elsif required.zero?
-        "expected at most #{times(maximum)}"
+        "expected at most #{count(maximum)}"
       else
-        "expected between #{required} and #{times(maximum)}"
+        "expected between #{required} and #{count(maximum)}"
       end
     end
     # rubocop:enable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
 
     def invoked_times
-      "invoked #{times(@invocations.size)}"
+      "invoked #{count(@invocations.size)}"
     end
 
     def actual_invocations
@@ -87,7 +87,7 @@ module Mocha
 
     attr_reader :required, :maximum
 
-    def times(number)
+    def count(number)
       case number
       when 0 then 'never'
       when 1 then 'once'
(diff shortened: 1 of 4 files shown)


test/acceptance/expected_invocation_count_test.rb
```diff
@@ -226,4 +226,15 @@ def test_should_fail_fast_if_there_is_no_matching_expectation
...
+  def test_should_pass_if_cardinality_is_satisfied_using_calls_made_both_before_and_after_updating_cardinality
+    test_result = run_as_test do
+      mock = mock('mock')
+      expectation = mock.expects(:method)
+      mock.method
+      expectation.twice
+      mock.method
+    end
+    assert_passed(test_result)
+  end
```
floehopper: Excellent - thanks for adding an acceptance test. I think it does make sense to make this behaviour part of the public API.

lib/mocha/cardinality.rb
```diff
@@ -96,6 +93,12 @@ def count(number)
       end
     end
 
+    def quantify(required, maximum)
+      @required = required
+      @maximum = maximum
+      self
```
floehopper: Do we definitely need to return `self` here? The method chaining is all at the `Expectation` level, so I would've thought it would be the methods on that class which would return `self`, no?

lib/mocha/expectation.rb
```diff
@@ -149,7 +149,6 @@ def at_least(minimum_number_of_times)
     #   # => verify fails
     def at_least_once
       at_least(1)
-      self
```
floehopper: As per my earlier comment, I think we might want to leave these methods returning `self`. While the calls to cardinality-related methods on `Expectation` are often the last in a chain, I suspect that isn't always the case. Also it feels inconsistent to only make this change to these two cardinality-related methods and not the others, but maybe I'm missing something.

Perhaps this highlights some missing acceptance tests...?

lib/mocha/cardinality.rb
```diff
@@ -87,7 +87,7 @@ def actual_invocations
 
     attr_reader :required, :maximum
 
-    def times(number)
+    def count(number)
```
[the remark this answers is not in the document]
floehopper: Thanks - this is now much clearer. 👍

test/acceptance/expected_invocation_count_test.rb
floehopper: > To me actually, it only _kinda_ makes sense. Updating expectations after the SUT has already been (partly - or even worse, fully) exercised seems like an odd thing to do that should, in theory, be disallowed or discouraged. I'm not sure if that's possible or desirable in practice, though.
> 
> So, pragramatically, this seems like the _most_ sensible even if not the _fully_ sensible choice.

I see what you mean. However, I don't feel strongly enough that it should be disallowed or even discouraged, so I'm happy to stick with what you've done in this PR.

lib/mocha/cardinality.rb
floehopper: I see. I read the code too quickly and assumed that returning `self` in this method was connected to the changes to the methods in `Expectation`, but I now realise they are not connected.

lib/mocha/expectation.rb
floehopper: Sorry. I'd completely missed that these methods call other methods on `Expectation` which themselves already return `self`. In my haste I somehow assumed they were calling methods on `Cardinality`.

lib/mocha/expectation.rb
```diff
@@ -549,7 +547,7 @@ def initialize(mock, expected_method_name, backtrace = nil)
       @block_matcher = BlockMatchers::OptionalBlock.new
       @ordering_constraints = []
       @side_effects = []
-      @cardinality = Cardinality.exactly(1)
+      @cardinality = Cardinality.new.exactly(1)
```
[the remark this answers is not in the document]
floehopper: Cool 👍

floehopper:
@nitishr Thanks again for identifying the problem in #473 and for taking the time to submit this PR and address my comments.

> However, our current rule is: Every cardinality-setting method defines both lower and upper bounds, thus replaces/overrides any earlier cardinality wholesale. This might even be worth documenting.

Agreed. I've opened #476 to address this.

> Most such 'combinations' yield no sensible/intuitive outcome. So, I think rather than supporting 'combinations', we should keep it simple and stick with 'overrides'.

Agreed. Thanks for considering my suggestion and for pointing out the flaws in it!

I'll try to get this merged shortly.

lib/mocha/cardinality.rb
```diff
@@ -96,6 +93,12 @@ def count(number)
       end
     end
 
+    def quantify(required, maximum)
```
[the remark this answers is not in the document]
floehopper: I'm still not convinced about the name of this method, but I'm happy enough to merge this PR as it stands and address the method name separately if necessary.

lib/mocha/expectation.rb
```diff
@@ -549,7 +549,7 @@ def initialize(mock, expected_method_name, backtrace = nil)
       @block_matcher = BlockMatchers::OptionalBlock.new
       @ordering_constraints = []
       @side_effects = []
-      @cardinality = Cardinality.exactly(1)
+      @cardinality = Cardinality.new(1, 1)
```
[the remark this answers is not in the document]
floehopper: This comment is against an outdated commit and it has been addressed elsewhere. Marking as resolved.

PR #477, opened 2020-03-15
API docs: avoid warning

olleolleolle:
This PR fixes a YARD annotation issue which hid the name of the param in the rendered API docs.

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -486,7 +486,7 @@ module Mocha
 
     # Constrains the expectation to occur only when the +state_machine+ is in the state specified by +state_predicate+.
     #
-    # @param [StateMachine::StatePredicate] state_predicate state_machine.is(state_name) provides a mechanism to determine whether the +state_machine+ is in the state specified by +state_predicate+ when the expected method is invoked.
+    # @param state_predicate [StateMachine::StatePredicate] +state_machine.is(state_name)+ provides a mechanism to determine whether the +state_machine+ is in the state specified by +state_predicate+ when the expected method is invoked.
     # @return [Expectation] the same expectation, thereby allowing invocations of other {Expectation} methods to be chained.
     #
     # @see API#states


floehopper:
@olleolleolle 

Thanks for submitting this PR.

This is what I see before the change:

before

And this is what I see after the change:

after

The only change I can see is that `state_machine.is(state_name)` is now formatted as code in red. Is that the only change you see too? Or was the parameter name, `state_predicate` actually invisible for you?

If you do see something different, can you be more specific about how the name of the param was hidden? Can you screenshot it? Where did you see it? What browser where you using?

Another thing I noticed in your PR was that you moved the parameter name to before the type declaration. Can you explain why you did that?

I looked up the YARD docs for a `@param` tag and was surprised to see that the order should be: [`@param name [Types] description`][1]. I noticed that *all* of the Mocha YARD docs use `@param [Types] name description` which seems odd. I trawled through the historical YARD documentation, CHANGELOG, and issues and [eventually discovered][2] that although the documented order is recommended, the (legacy) order that Mocha is using is still supported - phew!

Anyway, I've created [a modified version of your PR][3] where I just add the code block wrapping, i.e. don't use the new `@param` style. As far as I can see, this achieves the same effect, but means the code is more consistent with the other Mocha YARD docs. Does that make sense to you?

[1]: https://www.rubydoc.info/gems/yard/0.9.24/file/docs/Tags.md#param
[2]: https://github.com/lsegal/yard/issues/937
[3]: https://github.com/freerange/mocha/compare/yardfix

floehopper:
While investigating this I also noticed that YARD generates HTML including obsolete teletype tags and so I've opened [an issue against the YARD repo][1]!

[1]: https://github.com/lsegal/yard/issues/1316

olleolleolle:
The YARD warning I had originally was:

> lib/mocha/expectation.rb:494: [UnknownParam] @param tag has unknown parameter name: state_machine.is(state_name)

And, that's _not_ what we want. The new-style YARD `@param` usage (name first, then type specification, then description) avoids that.

floehopper:
Thanks for the quick response.

> The YARD warning I had originally was:
> 
> > lib/mocha/expectation.rb:494: [UnknownParam] @param tag has unknown parameter name: state_machine.is(state_name)

Hmm. That's odd. What version of YARD are you using? What YARD command did you run? And what version of the Mocha code did you run it against?

I'm using the latest version (v0.9.24) and I don't see any warnings when running on the `master` branch:

```
$ MOCHA_GENERATE_DOCS=1 GOOGLE_ANALYTICS_WEB_PROPERTY_ID=123 rake generate_docs
Files:          35
Modules:        11 (    4 undocumented)
Classes:        30 (    0 undocumented)
Constants:       0 (    0 undocumented)
Attributes:      1 (    0 undocumented)
Methods:        74 (    0 undocumented)
 96.55% documented
Updated 1 path from the index
```

> And, that's _not_ what we want. The new-style YARD `@param` usage (name first, then type specification, then description) avoids that.

If I can reproduce the warning, I'm happy to make the change, but if so I might want to make the order change for *all* `@param` tags across the Mocha codebase in one go.

> My reason given was probably NOT something I experienced, but later rationalized.
> 
> > which hid the name of the param

Ah, OK. That makes sense.

olleolleolle:
Ah, I used `yard doc` Rake task, not the specific `MOCHA_GENERATE_DOCS=1 GOOGLE_ANALYTICS_WEB_PROPERTY_ID=123 rake generate_docs` one.

And this was with the gem, not `master`.

olleolleolle:
So, I'll close this issue, and thank you for opening a YARD improvement one.

floehopper:
OK - thanks! I'm going to merge the other bit of your change anyway, because I think that's an improvement.

floehopper:
I've merged the other bit of the change in b4f59daa69a29b2ef1f9dae4160c4aa26898e4e4.

Thinking about it a bit more. Your usage seems pretty reasonable, i.e. wanting to be able to run `yard doc` against the gem code. The rake task is only really intended to be used in publishing [the online docs](https://mocha.jamesmead.org/), so I've opened [a new issue](https://github.com/freerange/mocha/issues/479) to ensure that `yard doc` works OK.

PR #496, opened 2020-11-01
Add note and example for yield with no args

### LINKED ISSUE #495: Stub and yield passed block
I was looking for way to stub a method and simply yield the block.

After trying some awkward approaches, I discovered this works with simply `stubs(:foo).yields`, but this form without arguments doesn't seem be documented:

https://www.rubydoc.info/github/floehopper/mocha/Mocha%2FExpectation:yields

If this is a valid use then I can submit a PR to update the docs.

andyw8:
Closes #495

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -265,6 +265,8 @@ module Mocha
     #
     # May be called multiple times on the same expectation for consecutive invocations.
     #
+    # If no +parameters+ are given then it will yield the block provided.
+    #
     # @param [*Array] parameters parameters to be yielded.
     # @return [Expectation] the same expectation, thereby allowing invocations of other {Expectation} methods to be chained.
     # @see #then
@@ -292,6 +294,13 @@ module Mocha
     #   sum # => 1
     #   fibonacci.next_pair { |first, second| sum = first + second }
     #   sum # => 2
+    #
+    # @example Yield the provided block
+    #   object = mock()
+    #   object.expects(:log).yields
+    #   messages = ""
+    #   object.log { messages += "log called" }
+    #   messages # => "log called"
     def yields(*parameters)
       multiple_yields(parameters)
     end


floehopper:
@andyw8 Thanks for opening this PR. I'm afraid I'm really busy at the moment, but I promise I'll get to it as soon as I can!

lib/mocha/expectation.rb
```diff
@@ -265,6 +265,8 @@ def with_no_block_given
     #
     # May be called multiple times on the same expectation for consecutive invocations.
     #
+    # If no +parameters+ are given then it will yield the block provided.
```
nitishr, line 268: This appears incorrect to me. This is just a special case of
> it yields with the specified +parameters+

from https://github.com/freerange/mocha/pull/496/files#diff-4c63dbab29ecbfb14e097947e3e0b295ad1382843978026979062d85b5287288R264

If we wanted to be more explicit about it, something like this would be more accurate
```suggestion
    # If no +parameters+ are given then it will yield to the block provided without any parameters.
```

This also makes me think, BTW, that perhaps there's a need to make the [introductory comment](https://github.com/freerange/mocha/pull/496/files#diff-4c63dbab29ecbfb14e097947e3e0b295ad1382843978026979062d85b5287288R264) clearer. Something like:

> Modifies expectation so that when the expected method is called, it yields with the specified +parameters+ to the block provided when calling. If no block is provided at the time of calling, the method will still yield, resulting in a +LocalJumpError+, as is standard ruby behavior.

lib/mocha/expectation.rb
```diff
@@ -292,6 +294,13 @@ def with_no_block_given
     #   sum # => 1
     #   fibonacci.next_pair { |first, second| sum = first + second }
     #   sum # => 2
+    #
+    # @example Yield the provided block
```
nitishr, line 298: Similar to above, this would be more accurate:

```suggestion
    # @example Yield to the block provided when calling
```

But then, how is this different from the [first example](https://github.com/freerange/mocha/pull/496/files#diff-4c63dbab29ecbfb14e097947e3e0b295ad1382843978026979062d85b5287288R274-R280)? This example might be simpler than the other one, being focused only on yielding, rather than yielding followed by returning as in the first example. So, we can consider replacing the first one with this example. In essence, though, they appear to be the same to me.

Am I missing something?

lib/mocha/expectation.rb
floehopper: For comparison purposes, I'm copying the current version of the introductory documentation in here:

> Modifies expectation so that when the expected method is called, it yields with the specified +parameters+ (even if no block is provided, in which case yielding will result in a +LocalJumpError+).

floehopper: @nitishr I agree with both your suggested changes. Thanks.

lib/mocha/expectation.rb
floehopper: @nitishr I agree with your suggested change to the wording - thanks. As I mentioned elsewhere, I'm happy to add more examples or simplify/split the existing ones if it makes things clearer.

floehopper:
@andyw8 Sorry for the very slow response - I've just been really busy with work. I've tweaked the changes in this PR a bit and merged them in d59081262f64a872d6ec5ec9dc9c0d7be57b643a. Closing.

andyw8:
Thanks!

floehopper:
Apologies - I merged this into the wrong branch. I've now added it into `main` in [this commit][1].

[1]: https://github.com/freerange/mocha/commit/1b6571c6c025d938fb92f16df997c8156f0739eb

PR #513, opened 2021-06-12
Fix misleading exception in `has_entry`

cstyles:
If `has_entry` is called without any arguments, it will raise an
exception like the following:

```
ArgumentError: Too many arguments; use either a single argument (must be a Hash) or two arguments (a key and a value).
```

This is because we don't check if the `options.length` is zero so when
it is, the `else` branch executes and prints the above error. We can fix
this by simply adding another branch to cover the case when
`options.length` is zero.

---

Unfortunately, this change introduces a new RuboCop violation:

```
$ bundle exec rake test
Running RuboCop...
Inspecting 234 files
................................................................................................................................................................................
............................................C.............

Offenses:

lib/mocha/parameter_matchers/has_entry.rb:43:5: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for has_entry is too high. [7/6]
    def has_entry(*options) # rubocop:disable Naming/PredicateName ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

234 files inspected, 1 offense detected
RuboCop failed!
```

We could disable the cop but I didn't feel like it was my call to make. I'll leave that up to y'all.

### DIFF
--- a/lib/mocha/parameter_matchers/has_entry.rb
+++ b/lib/mocha/parameter_matchers/has_entry.rb
@@ -42,6 +42,8 @@ module Mocha
     #
     def has_entry(*options) # rubocop:disable Naming/PredicateName
       case options.length
+      when 0
+        raise ArgumentError, 'Argument has no entries.'
       when 1
         case options[0]
         when Hash
--- a/test/unit/parameter_matchers/has_entry_test.rb
+++ b/test/unit/parameter_matchers/has_entry_test.rb
@@ -92,6 +92,11 @@ class HasEntryTest < Mocha::TestCase
     assert_equal 'Argument has no entries.', e.message
   end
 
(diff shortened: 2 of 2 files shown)


floehopper:
@cstyles Many thanks for this. It looks good to me. I'm having some issue with the Travis CI builds at the moment and I'm a bit short of time, so I might not get it merged for a few days. Please bear with me.

floehopper:
I've fixed this up a bit and merged it in 9c4ef13197300d6edd1990dabd3df00385e40d98. Thanks again for this contribution. 😄

floehopper:
Released in v1.13.0.

See https://mocha.jamesmead.org/Mocha/ParameterMatchers.html#has_keys-instance_method.

PR #519, opened 2021-11-11
Allow hiding satisfied expectations in test output

thomasmarshall:
It's often not important to see satisfied stubs in test output, and in some test suites the noise can be difficult to parse. This commit introduces an `Expectation#quietly` method which allows the suppression of test output for satisfied expectations on a case-by-case basis.

### Example

```rb
setup do
  SomeGlobal.stubs(:foo).with(any_of(*SomeGlobal::THINGS)).quietly
  AnotherGlobal.stubs(:connection).quietly
end

test "foo" do
  Foo.stubs(:bar)
  Bar.expects(:baz)
end
```

| Without `#quietly` | With `#quietly` |
| - | - |
| Screenshot 2021-11-11 at 15 35 17 | Screenshot 2021-11-11 at 17 46 29 |

### Alternatives

One alternative might be to treat stubs and expectations differently, such that satisfied stubs are never reported. Or maybe satisfied expectations with a cardinality of 0 needn't be reported. Or maybe it's already possible to hook into `Mockery` somehow to override the reporting behaviour?

At the moment, we're monkey patching to achieve something similar but thought others might be interested in this change – no worries if not!

### DIFF
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -541,6 +541,18 @@ module Mocha
       self
     end
 
+    # Modifies expectation so that it will not be reported in test output, if it is satisfied.
+    #
+    # @return [Expectation] the same expectation, thereby allowing invocations of other {Expectation} methods to be chained.
+    #
+    # @example Satisfied expectation will not be reported in test output.
+    #   object = mock()
+    #   object.stubs(:expected_method).quietly
+    def quietly
+      @quiet = true
+      self
+    end
+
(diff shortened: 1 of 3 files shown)


floehopper:
@thomasmarshall Thanks for the PR. At first glance doing something about this sounds sensible. I wonder whether a configuration option might be more appropriate. I'm on leave at the moment, but I'll try to have a proper look at it when I'm back next week.

PR #527, opened 2022-04-22
Move development dependencies from gemspec to Gemfile

mishina2228:
Gemspec should not contain `RUBY_VERSION` as a condition to switch dependencies,
because it is meaningless.
Please see the following for details:
- https://github.com/rubocop/rubocop/issues/7137
- https://github.com/rubocop/ruby-style-guide/issues/763

In addition, Bundler no longer uses `add_development_dependency` when genererating new gems.
- https://github.com/rubygems/bundler/pull/7222

For these reason, this PR moves development dependencies from gemspec to Gemfile.

 The CI status can be seen here: https://app.circleci.com/pipelines/github/mishina2228/mocha/9/workflows/2c6bbae9-c2c8-44f7-bf86-2f0fb487b83d

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,34 @@
 source 'https://rubygems.org'
 
 gemspec
+
+# rubocop:disable Bundler/DuplicatedGem
+if RUBY_VERSION < '1.9.3'
+  gem 'rake', '~> 10.0'
+elsif RUBY_VERSION < '2'
+  gem 'rake', '~> 12.2.1'
+elsif RUBY_VERSION < '2.2'
+  gem 'rake', '~> 12.3.3'
+else
+  gem 'rake'
+end
+# rubocop:enable Bundler/DuplicatedGem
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -1,3 +1,34 @@
 source 'https://rubygems.org'
 
 gemspec
+
+# rubocop:disable Bundler/DuplicatedGem
```
mishina2228, line 5: RuboCop v0.58.2 has a false positive for this case.
So I had to disable `Bundler/DuplicatedGem`.
The latest RuboCop (v1.27.0) raises no offense.

floehopper:
@mishina2228 This is great - thanks! I've tweaked your commit very slightly so the order of the gems in the two `Gemfile.*.latest` files are consistent. The changes have been merged in dd127f7bba4e0e58e96f687e352edc5e4f80f334. Closing.

Issue #531, opened 2022-08-15
Investigate rationalising Configuration.stubbing_non_existent_method= & Mock#responds_like


floehopper:
While investigating #149 I realised that there's some overlap between the behaviour controlled by [`Configuration.stubbing_non_existent_method=`](https://mocha.jamesmead.org/Mocha/Configuration.html#stubbing_non_existent_method=-instance_method) and the responder-related behaviour controlled by [`Mock#responds_like`](https://mocha.jamesmead.org/Mocha/Mock.html#responds_like-instance_method) & [`Mock#responds_like_instance_of`](https://mocha.jamesmead.org/Mocha/Mock.html#responds_like_instance_of-instance_method).

### The configuration check

* Is done at stubbing time.
* Uses both `ClassMethods#__method_exists__?` (with `include_public_methods` set to `true`) *and* `Object#respond_to?` (with `include_all` defaulting to `false`). 
* `ClassMethods#__method_exists__?` in turn uses `Module#public_method_defined?`, `Module#protected_method_defined?` & `Module#private_method_defined?` with `inherited` set to `true`.

### The responder check

* Is done at invocation time.
* Just uses `Object#respond_to?` with `include_all` set to `true`.

I'm not entirely sure why we can't just use `Object#respond_to?` with `include_all` set to `true` in both cases. It would be worth making this change and seeing if any (acceptance) tests fail.

If we can use the same check for both, it would then be worth considering only making the check in one place. My inclination is to do this at stubbing time which I think is what RSpec does. However, note that the current implementation in `Mockery.on_stubbing` only seems to be triggered for partial mocks (i.e. from `ObjectMethods#expects` & `ObjectMethods#stubs`), so we'd need to change that to make it work for mock objects with a responder.

Issue #149, opened 2013-04-15
Mocha::Configuration.prevent(:stubbing_non_existent_method) should check Method#arity / Method#parameters


floehopper:
And maybe method visibility?

floehopper:
Also see [`Method#parameters`](https://ruby-doc.org/core-2.7.0/Method.html#method-i-parameters).

floehopper:
> I did a bit of spiking on this approach - I think we'd need to introduce a hook within the method defined in `Mocha::ClassMethod#define_new_method`.This hook would be invoked when the stubbed method was invoked. In order to do this, I suspect we'd need to convert the `class_eval`/String code to use `define_method` something like what I've done in [this branch](https://github.com/freerange/mocha/tree/spike-check-invocation-of-stubbed-method-against-original-method-signature).

The changes in [the spike branch](https://github.com/freerange/mocha/tree/spike-check-invocation-of-stubbed-method-against-original-method-signature) seem to have been confined to [a single commit](https://github.com/freerange/mocha/commit/bd9e1c823c9efbe22c5dba561d5984fcbf4c9ae9). The changes in that commit were effectively implemented in [this commit](https://github.com/freerange/mocha/commit/8a8e30d60d0b055b801a66c4d79ac233478c26b0) some 6 years later! So if we did want to introduce a hook along the lines of my earlier comment, it ought to be possible but the code now lives in `StubbedMethod#define_new_method`:

https://github.com/freerange/mocha/blob/ed9c040a9550273e52e78c415de7f5bfbef992d4/lib/mocha/stubbed_method.rb#L57-L64

floehopper:
Unsurprisingly it looks as if RSpec's functionality has moved on considerably since [this comment](https://github.com/freerange/mocha/issues/149#issuecomment-27167077). In particular, verifying doubles now check the method signature using `Method#parameters` and not just `Method#arity`. Also much of this logic has been extracted into a new `rspec-support` gem, notably [`MethodSignatureVerifier`](https://github.com/rspec/rspec-support/blob/528d88ce6fac5f83390bf430d1c47608e9d8d29a/lib/rspec/support/method_signature_verifier.rb). The comment for `StrictSignatureVerifier` in this file made me laugh:

> Figures out wether a given method can accept various arguments. Surprisingly non-trivial.

floehopper:
[...]

PR #532, opened 2022-08-15
Implicitly set responder on partial mocks

floehopper:
This relates to #149 and #531.

All the tests seem to pass, but it would be worth doing a bit more
thinking about the change in Mock#check_responder_responds_to where we
now set include_all to true for the call to Object#respond_to?.

Also it's only really worth doing this if the investigation in #531
means that it's definitely worthwhile. It would probably also be worth
doing some spiking on the responder-related solution proposed in #149.


floehopper:
Note that the build failure on Ruby v1.8 looks like it's due to blocks not supporting argument default values. That was only introduced in Ruby v1.9.

floehopper:
> Note that the build failure on Ruby v1.8 looks like it's due to blocks not supporting argument default values. That was only introduced in Ruby v1.9.

Now fixed via #533.

floehopper:
Rebased now that #533 has been merged.

floehopper:
See also #580 & #583.

floehopper:
I'm not 100% sure about the changes and the commit note and PR description could be improved, but I think this is now at the point where it's ready for a proper review.

* [ ] Explain the changes in `RespondsLikeTest` in the commit note
* [ ] Update the PR description

floehopper:
Some work on #605 has highlighted a potentially serious flaw in this approach. Currently for partial mocks, the method obtainable from the responder is the stub method defined by Mocha and not the original method which is what we want. More thought required!

floehopper:
> Some work on #605 has highlighted a potentially serious flaw in this approach. Currently for partial mocks, the method obtainable from the responder is the stub method defined by Mocha and not the original method which is what we want. More thought required!

Closing on the basis that #531 is tracking the work on this.

PR #542, opened 2022-08-28
A few cleanups following removal of support for Ruby v1.8

nitishr:
... in #536

### DIFF
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -118,10 +118,9 @@ module Mocha
     end
 
     def on_stubbing(object, method)
-      method = method.to_sym
       signature_proc = lambda { "#{object.mocha_inspect}.#{method}" }
       check(:stubbing_non_existent_method, 'non-existent method', signature_proc) do
-        !(object.stubba_class.__method_exists__?(method, true) || object.respond_to?(method.to_sym))
+        !(object.stubba_class.__method_exists__?(method, true) || object.respond_to?(method))
       end
       check(:stubbing_non_public_method, 'non-public method', signature_proc) do
         object.stubba_class.__method_exists__?(method, false)
--- a/lib/mocha/stubbed_method.rb
+++ b/lib/mocha/stubbed_method.rb
@@ -73,7 +73,7 @@ module Mocha
     def restore_original_method
       return if use_prepended_module_for_stub_method?
       if stub_method_overwrites_original_method?
-        original_method_owner.send(:define_method, method_name, method_body(@original_method))
+        original_method_owner.send(:define_method, method_name, @original_method)
       end
       retain_original_visibility(original_method_owner)
     end
--- a/test/assertions.rb
+++ b/test/assertions.rb
@@ -1,6 +1,5 @@
 module Assertions
   def assert_method_visibility(object, method_name, visiblity)
-    method_key = method_name.to_sym
-    assert object.send("#{visiblity}_methods").include?(method_key), "#{method_name} is not #{visiblity}"
+    assert object.send("#{visiblity}_methods").include?(method_name.to_sym), "#{method_name} is not #{visiblity}"
   end
 end
--- a/lib/mocha/any_instance_method.rb
+++ b/lib/mocha/any_instance_method.rb
@@ -8,10 +8,6 @@ module Mocha
       stubbee.any_instance
     end
 
-    def method_body(method)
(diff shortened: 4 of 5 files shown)


test/assertions.rb
```diff
@@ -1,6 +1,5 @@
 module Assertions
   def assert_method_visibility(object, method_name, visiblity)
-    method_key = method_name.to_sym
-    assert object.send("#{visiblity}_methods").include?(method_key), "#{method_name} is not #{visiblity}"
+    assert object.send("#{visiblity}_methods").include?(method_name), "#{method_name} is not #{visiblity}"
```
floehopper, line 3: Is this possible, because `Object#public_methods`, etc, returns an array of strings in Ruby > v1.8; but returns an array of symbols in Ruby v1.8? If so, it would be good to explain that in the commit note.

lib/mocha/stubbed_method.rb
```diff
@@ -8,9 +8,9 @@ class StubbedMethod
 
     def initialize(stubbee, method_name)
       @stubbee = stubbee
+      @method_name = method_name
       @original_method = nil
       @original_visibility = nil
-      @method_name = method_name.to_sym
```
floehopper, line 13: How do you know it's safe to make this change and that it won't result in a change in behaviour under some circumstances? Why is it safe to change this call, but not other calls to `String#to_sym` in the rest of the codebase?

floehopper (changes requested):
Hi @nitishr. Many thanks for submitting this. The first commit looks good to me, but I'm a bit less sure about the 2nd one. I've added some questions inline.

lib/mocha/mockery.rb
```diff
@@ -118,10 +118,9 @@ def mocha_inspect
     end
 
     def on_stubbing(object, method)
-      method = method.to_sym
       signature_proc = lambda { "#{object.mocha_inspect}.#{method}" }
       check(:stubbing_non_existent_method, 'non-existent method', signature_proc) do
-        !(object.stubba_class.__method_exists__?(method, true) || object.respond_to?(method.to_sym))
+        !(object.stubba_class.__method_exists__?(method, true) || object.respond_to?(method))
```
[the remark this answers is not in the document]
nitishr: More or less. I've pushed a new commit with an explanation.

test/assertions.rb
nitishr: Yes. And existing callers of `assert_method_visibility` only ever pass the `method_name` as a symbol.

However, on second thoughts, instead of removing the `to_sym` conversion, I've just inlined the temp `method_key`.

lib/mocha/stubbed_method.rb
nitishr: I was relying on existing tests to catch issues, but now I think it's better to leave this conversion in place here. For the most part, it should work either way since the `method_name` attr of `StubbedMethod` is ultimately gets passed to ruby library methods that'll accept either a symbol or a string (e.g. `define_method`, `remove_method`, etc.). However, `StubbedMethod#matches` might behave differently if one instance has the `method_name` as string and the other has it as a symbol.

Hence, I've undone this change.

lib/mocha/mockery.rb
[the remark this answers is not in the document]
floehopper: Thanks. I've done a bit more investigation and realised that this change could have been made prior to dropping support for Ruby v1.8. I've updated the commit note accordingly.

lib/mocha/stubbed_method.rb
floehopper: Thanks. I know I'm probably being excessively cautious and I know I should rely on the tests to make these kinds of changes, but I've been burned so many times by people relying on undocumented behaviour that I'd like to be more confident that a change like this isn't going to break something accidentally.

test/assertions.rb
floehopper: Thanks. I've convinced myself that not calling `to_sym` is safe here, so I've made that change.

floehopper (commented):
I belatedly realised that it is safe to remove the call to `to_sym` from `Assertions#assert_method_visibility`, so I've changed the last commit to do that. Since strictly speaking not all these changes were enabled by the dropping of support for Ruby v1.8, I've merged these commit manually rather than grouping them into a merge commit.

floehopper:
Merged manually. Closing.

Issue #537, opened 2022-08-21
Use Ruby v1.9 Hash literal syntax


floehopper:
Now that we've dropped support for Ruby v1.8 in #536, we should be able to configure the `Style/HashSyntax` cop to use a more modern style across the board.

floehopper:
The fix for this was released in v2.0.3.

Issue #446, opened 2019-12-17
Work out whether any changes are needed for positional/keyword args in Ruby 3.0


floehopper:
See: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

wasabigeek:
I'm having trouble matching keyword arguments precisely in Ruby 3. The problem seems to be that the `expected_parameters` look the same for both these cases:
- `.with("example", { foo => "bar" })`  => expected parameters are `[“example”, { foo => “bar” }]`
- `.with("example", foo => "bar")` (or `.with("example", **{ foo => "bar"})`) => expected parameters are also `[“example”, { foo => “bar” }]`

But they aren't the same in Ruby 3. For example:

```ruby
require "minitest/autorun"
require 'mocha/minitest'
class Example
  def splatted_foo(*args, **kwargs)
    modified_kwargs = kwargs.merge({ example: :text })
    bar(*args, **modified_kwargs)
  end

  def nonsplatted_foo(*args, **kwargs)
    modified_kwargs = kwargs.merge({ example: :text })
    bar(*args, modified_kwargs)
  end

  def bar(*args, hello:, example:)
  end
end

class Test < Minitest::Test
  def test_wrapper_passes_arguments
    example = Example.new
    args = [1]
    kwargs = {hello: :world}
    expected_kwargs = {hello: :world, example: :text}
    example.expects(:bar).with(1, expected_kwargs) # passes
    example.splatted_foo(*args, **kwargs)

    example.expects(:bar).with(1, expected_kwargs) # passes
    example.nonsplatted_foo(*args, **kwargs) # however, this same code would raise an ArgumentError if run in console
  end
end
```
Above was tested with ruby 3.0.3, minitest (5.15.0), mocha (1.13.0).

floehopper:
@wasabigeek Thanks for reporting this. I'm on holiday at the moment, so I might not be able to look at this properly for a while. Just to make sure I understand, can you confirm that the main issue is that the stubbed version of `Example#foo` doesn't behave the same way as the real version of the method, because the stubbed version does not raise an `ArgumentError` in cases where the real method would do so...?

wasabigeek:
> the main issue is that the stubbed version of Example#foo doesn't behave the same way as the real version of the method, because the stubbed version does not raise an ArgumentError in cases where the real method would do so...?

[...]

PR #547, opened 2022-09-02
Use Ruby v1.9 features

nitishr:
Slightly expanding on #537, now that we've dropped support for Ruby v1.8 in https://github.com/freerange/mocha/pull/536, we should be able to enable the Style/HashSyntax, Style/EachWithObject and Style/FormatStringToken cops across the board.

The first two commits include just the output of rubocop autocorrect.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,10 +19,6 @@ Style/ExpandPathArguments:
 Style/FormatStringToken:
   EnforcedStyle: unannotated
 
-# ruby19 style has only been supported since Ruby v1.9
-Style/HashSyntax:
-  EnforcedStyle: hash_rockets
-
 # I'm not keen on this cop, because it's easy to miss the conditional
 # I think the results are particularly unhelpful when Metrics/LineLength is big
 Style/IfUnlessModifier:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,10 +15,6 @@ Style/Documentation:
 Style/ExpandPathArguments:
   Enabled: false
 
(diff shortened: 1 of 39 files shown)


floehopper:
@nitishr

Thanks for submitting this PR. I've just belatedly enabled pull request builds on the CircleCI. Can you bump this branch to kick off a CircleCI build?

Could you separate the final "DRY up mocha_inspect" commit into a separate PR? I understand it depends on the string formatting changes, but otherwise it feels orthogonal to this PR and I'd prefer to consider it separately. We can always apply the string formatting changes to it separately.

As regards the other changes, I'd be happy to merge the two smaller commits, but I'm a bit hesitant about the `Style/HashSyntax` one, because (since it's quite pervasive) it might make life difficult for the ongoing work on supporting better matchers for keyword arguments in #446, #543 & #544. Since the changes were generated using Rubocop auto-correct, it might just be simpler to wait for the keyword matching changes to land before then re-running the auto-correct at that point.

Having said all that, I can see why it would be nice to merge the 3 commits together.

I'm afraid I've got very limited time to work on Mocha at the moment and I'm keen to focus the time I do have on the changes related to keyword argument matching which provide new/improved functionality to users. I hope that makes sense.

floehopper:
Closed by 26c4ae1ecd1057751685a5474932d05701a2c294.

floehopper:
Oh, on second thoughts maybe this isn't actually closed - presumably this relates to features available in Ruby >= v1.9 and not just Ruby v1.9! Re-opening.

Issue #325, opened 2018-03-16
Consider dropping support for older versions of Ruby


floehopper:
Ideally come up with rules to define how this works.

Some useful references:

* https://github.com/rspec/rspec/issues/25
* https://github.com/rspec/rspec-rails/pull/1967

floehopper:
[Extended maintenance of Ruby versions 1.8.7 and 1.9.2 will end on July 31, 2014](https://www.ruby-lang.org/en/news/2014/07/01/eol-for-1-8-7-and-1-9-2/).

nitishr:
Why aren't we dropping support for at least 1.8.7 already? It hasn't even been possible for me to run `bundle exec rake` with 1.8.7 due to the difficulty of installing the right combination of gems and libraries.

This would also help clean up some code as well as close some open issues.

floehopper:
> Why aren't we dropping support for at least 1.8.7 already?

No good reason. I think I was just agonising over how to deal with it within the confines of semantic versioning. I've just done this in c5f8496d5102445de315a10efc385209a962bdc1.

> It hasn't even been possible for me to run bundle exec rake with 1.8.7 due to the difficulty of installing the right combination of gems and libraries.

That's odd. Did you try deleting your `Gemfile.lock` before running `bundle install`? That seems to work for me locally.

> This would also help clean up some code as well as close some open issues.

I agree. Thanks for the prompt!

floehopper:
[Extended maintenance of Ruby v1.8.7 ended on 31 Jul 2014][1] and Mocha support for Ruby v1.8 has been [deprecated since v1.10][2]. Recent work in #534 has run into problems with files containing keyword arguments not being parsable by Ruby v1.8 and so I think the time has come to drop support.

[1]: https://www.ruby-lang.org/en/news/2014/07/01/eol-for-1-8-7-and-1-9-2/
[2]: https://github.com/freerange/mocha/commit/c5f8496d5102445de315a10efc385209a962bdc1

floehopper:
I've just dropped support for Ruby v1.8 in #536. There's still some question about whether that should be a minor or major version release, but that decision can wait for now.

floehopper:
[Support for Ruby version 1.9.3 ended on February 23, 2015](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/).

floehopper:
[...]

PR #555, opened 2022-09-20
Deprecate support for Ruby versions earlier than v2.0

floehopper:
This was based on #553.

This is long overdue. [Extended maintenance of Ruby v1.9.3 ended on 23 Feb 2015][1]. Actually dropping support will allow significant simplification of the code.

The immediate motivation is to come up with better parameter matching for keyword arguments and we've run into parsing issues when trying to do this in conjunction with continuing to support Ruby v1.9.

See #325.

[1]: https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

### DIFF
--- a/lib/mocha/api.rb
+++ b/lib/mocha/api.rb
@@ -1,3 +1,4 @@
+require 'mocha/ruby_version'
 require 'mocha/parameter_matchers'
 require 'mocha/hooks'
 require 'mocha/mockery'
--- a/lib/mocha/minitest.rb
+++ b/lib/mocha/minitest.rb
@@ -1,3 +1,4 @@
+require 'mocha/ruby_version'
 require 'mocha/integration/mini_test'
 require 'mocha/deprecation'
 
--- a/lib/mocha/ruby_version.rb
+++ b/lib/mocha/ruby_version.rb
@@ -1,3 +1,11 @@
+require 'mocha/deprecation'
+
 module Mocha
   RUBY_V2_PLUS = Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2')
+
+  unless RUBY_V2_PLUS
+    Mocha::Deprecation.warning(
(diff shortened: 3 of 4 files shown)


floehopper:
@wasabigeek While I was expanding the commit note, I noticed that the deprecation warning wasn't showing up in the CI build for Ruby v1.9. This seems to be because `lib/mocha.rb` is not automatically required when the `mocha` gem is included in a `Gemfile`. Moving the deprecation warning into `lib/mocha/ruby_version.rb` seems to fix the problem and I can now see the warning in [the build output](https://app.circleci.com/pipelines/github/freerange/mocha/165/workflows/05b908a1-d096-4dc0-a403-558e18ec0a3d/jobs/2322/parallel-runs/0/steps/0-108). I suspect that means that the previous deprecation warning may not have actually worked correctly, although perhaps something else has changed in the meantime. Anyway, I'm pretty tired, so I'd appreciate another pair of eyes on this before I merge it!

wasabigeek (approved):
This works for me, though I couldn't figure out how to install Ruby 1.9 and had to reverse this conditional instead 😄.

If I understand correctly though, this comes up more as a side effect of the chain of `require`s kicked off by `require 'mocha/mini_test'` (or similar) i.e. if somebody somehow manages to initiate a code path that doesn't `require 'mocha/ruby_version'`, they won't see the warning. I guess this probably won't be an issue given that `mocha/api` is the common entry-point and it will trigger the warning 👍

floehopper:
@wasabigeek Thanks for looking at this - that's really helpful. I think I might try moving the deprecation warning into `mocha/api`. That makes more sense to me now.

floehopper:
In the end I've decided to leave the deprecation warning in 'mocha/ruby_version', but to require that file from 'mocha/test_unit', 'mocha/minitest' & 'mocha/api' to cater for Test::Unit, Minitest & RSpec integration respectively.

PR #535, opened 2022-08-20
Ruby3 keyword argument matching

floehopper:
TODO:

* [ ] Is there a missing unit test scenario in `ExpectationTest`?
* [ ] Add acceptance tests, c.f. [`ParameterMatcherTest`](https://github.com/freerange/mocha/blob/b7a7d23321e409c9ad77229f80033a55eb0d8a90/test/acceptance/parameter_matcher_test.rb) and/or [`OptionalParameterMatcherTest`](https://github.com/freerange/mocha/blob/b7a7d23321e409c9ad77229f80033a55eb0d8a90/test/acceptance/optional_parameters_test.rb).
* [ ] Work out whether [the two test failures in `StubbaExampleTest`](https://app.circleci.com/pipelines/github/freerange/mocha/128/workflows/59b1abe9-2659-4abf-b42f-797c62aeee07/jobs/1795) is a sign that this change might break a lot of existing tests. Do we need to put this change behind a configuration option?
* [x] ~Simplify/centralize Ruby version checking for ruby2 keywords functionality.~ Fixed by using `RUBY_V3_PLUS`.
* [x] ~Work out what to do with Ruby v1.8 - there is a problem *parsing* `expectation_test.rb` with keyword arguments.~ Resolved in #536
* [x] ~Resolve Rubocop issues in unit tests.~ Resolved by disabling `Style/BracesAroundHashParameters` cop and using hash rocket syntax in combination with `Hash.ruby2_keywords_hash` to designate keyword arguments. Now that we've dropped support for Ruby v1.8, we could probably change the configuration of the `Style/HashSyntax` cop, but that feels like a bigger change (see #537).
* [x] ~Fix Yardoc for `LastPositionalHash`.~ Fixed by marking the whole class as private from a documentation point of view. 
* [ ] Find someone with a large Rails codebase using Mocha to test a pre-release version

PR #562, opened 2022-10-01
Support strict keyword argument matching

### LINKED ISSUE #446: Work out whether any changes are needed for positional/keyword args in Ruby 3.0

floehopper:
This introduces a new `strict_keyword_argument_matching` configuration option. This option is only available in Ruby >= v2.7 and is disabled by default to enable gradual adoption.

When the strict keyword argument option is enabled, an expectation expecting keyword arguments (via `Expectation#with`) will no longer match an invocation passing a positional Hash argument.

Without this option, positional hash and keyword arguments are treated the same during comparison, which can lead to false negatives in Ruby >= v3.0 (see examples below).

### Loose keyword argument matching (default)

```ruby
class Example
  def foo(a, bar:); end
end

example = Example.new
example.expects(:foo).with('a', bar: 'b')
example.foo('a', { bar: 'b' })
# This passes the test, but would result in an ArgumentError in practice
```

### Strict keyword argument matching

```ruby
Mocha.configure do |c|
  c.strict_keyword_argument_matching = true
end

class Example
  def foo(a, bar:); end
end

example = Example.new
example.expects(:foo).with('a', bar: 'b')
example.foo('a', { bar: 'b' })
# This now fails as expected
```

For more details on keyword arguments in Ruby v3, refer to [this article][1].

Note that Hash matchers such as has_value or has_key will still treat positional hash and keyword arguments the same, so false negatives are still possible when they are used.

Closes #446.

See also #535 & #544 for discussions relating to this change.

[1]: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0

### DIFF
--- a/test/acceptance/keyword_argument_matching_test.rb
+++ b/test/acceptance/keyword_argument_matching_test.rb
@@ -20,6 +20,19 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
     assert_passed(test_result)
   end
 
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_hash_parameter_with_keyword_args_when_strict_keyword_matching_is_enabled
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(:key => 42)
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+        end
+      end
+      assert_failed(test_result)
+    end
+  end
+
   def test_should_match_hash_parameter_with_splatted_keyword_args
     test_result = run_as_test do
       mock = mock()
@@ -27,6 +40,19 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
       mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
     end
     assert_passed(test_result)
+  end
+
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_hash_parameter_with_splatted_keyword_args_when_strict_keyword_matching_is_enabled
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(**{ :key => 42 })
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+        end
+      end
+      assert_failed(test_result)
+    end
   end
 
   def test_should_match_splatted_hash_parameter_with_keyword_args
@@ -56,6 +82,19 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
     assert_passed(test_result)
   end
 
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_positional_and_keyword_args_with_last_positional_hash_when_strict_keyword_args_is_enabled
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method(1, :key => 42)
+        end
+      end
+      assert_failed(test_result)
+    end
+  end
+
   def test_should_match_last_positional_hash_with_keyword_args
     test_result = run_as_test do
       mock = mock()
@@ -63,6 +102,19 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
       mock.method(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
     end
     assert_passed(test_result)
+  end
+
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_last_positional_hash_with_keyword_args_when_strict_keyword_args_is_enabled
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(1, :key => 42)
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+        end
+      end
+      assert_failed(test_result)
+    end
   end
 
   def test_should_match_positional_and_keyword_args_with_keyword_args
@@ -108,5 +160,18 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
       mock.method(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
     end
     assert_passed(test_result)
+  end
+
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_non_hash_args_with_keyword_args
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(**{ :key => 1 })
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method([2])
+        end
+      end
+      assert_failed(test_result)
+    end
   end
 end
--- a/test/acceptance/keyword_argument_matching_test.rb
+++ b/test/acceptance/keyword_argument_matching_test.rb
@@ -1,0 +1,112 @@
+require File.expand_path('../acceptance_test_helper', __FILE__)
+
+class KeywordArgumentMatchingTest < Mocha::TestCase
+  include AcceptanceTest
+
+  def setup
+    setup_acceptance_test
+  end
+
+  def teardown
+    teardown_acceptance_test
+  end
+
+  def test_should_match_hash_parameter_with_keyword_args
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(:key => 42)
+      mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_hash_parameter_with_splatted_keyword_args
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(**{ :key => 42 })
+      mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_splatted_hash_parameter_with_keyword_args
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(:key => 42)
+      mock.method(**{ :key => 42 })
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_splatted_hash_parameter_with_splatted_hash
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(**{ :key => 42 })
+      mock.method(**{ :key => 42 })
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_positional_and_keyword_args_with_last_positional_hash
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+      mock.method(1, :key => 42)
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_last_positional_hash_with_keyword_args
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(1, :key => 42)
+      mock.method(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_positional_and_keyword_args_with_keyword_args
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(1, :key => 42)
+      mock.method(1, :key => 42)
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_hash_parameter_with_hash_matcher
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(has_key(:key))
+      mock.method({ :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_splatted_hash_parameter_with_hash_matcher
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(has_key(:key))
+      mock.method(**{ :key => 42 })
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_positional_and_keyword_args_with_hash_matcher
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(1, has_key(:key))
+      mock.method(1, :key => 42)
+    end
+    assert_passed(test_result)
+  end
+
+  def test_should_match_last_positional_hash_with_hash_matcher
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(1, has_key(:key))
+      mock.method(1, { :key => 42 }) # rubocop:disable Style/BracesAroundHashParameters
+    end
+    assert_passed(test_result)
+  end
+end
--- a/test/unit/parameter_matchers/positional_or_keyword_hash_test.rb
+++ b/test/unit/parameter_matchers/positional_or_keyword_hash_test.rb
@@ -1,0 +1,73 @@
+require File.expand_path('../../../test_helper', __FILE__)
+
+require 'mocha/parameter_matchers/positional_or_keyword_hash'
+require 'mocha/parameter_matchers/instance_methods'
+require 'mocha/inspect'
+
+class PositionalOrKeywordHashTest < Mocha::TestCase
+  include Mocha::ParameterMatchers
+
+  def test_should_describe_matcher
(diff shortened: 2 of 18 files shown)


floehopper:
> This looks good to me, but I found that the error output could be confusing as the last positional hashes / keyword arguments are not printed accurately. Should we address it in this PR? I'm also good to merge this and follow-up with a separate fix.
> 
> See [wasabigeek#2](https://github.com/wasabigeek/mocha/pull/2) for more details!

Thanks for getting back to me. Sorry for the delay in replying. I'll take a look at wasabigeek/mocha#2 now.

floehopper:
I've responded here: https://github.com/wasabigeek/mocha/pull/2#issuecomment-1272525903

floehopper:
> @floehopper, I've done a bit of testing in a larger repo at work + got some good feedback from a colleague:

Great - thank you!

> * hundreds of tests are breaking 😭 ideally it'd be _even_ easier to gradually adopt the config, e.g. add a variation of `#with` which will apply strict matching (such as `#with_strict_keywords`) even if the config is not enabled. I think that's a great idea, what do you think? IMO, it could be a separate PR.

Yes, I can see how that would be useful.

> * `Hash.ruby2_keywords_hash?` ([here](https://github.com/freerange/mocha/pull/562/files#diff-aeaab6126d1c65903c8dbc5de4cd0c4a9752d7afa1e6d05a631d5030951f3be9R15)) will raise an argument error if `parameter` is not a Hash - _this is a bug that we should address in this PR_. This is a bigger issue in a test where there are multiple expectations, since every expectation is re-checked against each invocation, e.g.:
>   ```ruby
>   example.expects(:foo).with(bar: 1) # expectation A
>   example.expects(:foo).with([1]) # expectation B
>   
>   example.foo(bar: 1)
>   example.foo([1]) # raises ArgumentError since expectation A is checked again (?), and Hash.ruby2_keywords_hash? raises an error since an Array is passed in
>   ```
>     
>   Adding a `return false unless parameter.is_a? Hash` should be sufficient.

OK. I'll take a look at fixing this now. Thanks.

floehopper:
> > * `Hash.ruby2_keywords_hash?` ([here](https://github.com/freerange/mocha/pull/562/files#diff-aeaab6126d1c65903c8dbc5de4cd0c4a9752d7afa1e6d05a631d5030951f3be9R15)) will raise an argument error if `parameter` is not a Hash - _this is a bug that we should address in this PR_. This is a bigger issue in a test where there are multiple expectations, since every expectation is re-checked against each invocation, e.g.:
> >   ```ruby
> >   example.expects(:foo).with(bar: 1) # expectation A
> >   example.expects(:foo).with([1]) # expectation B
> >   
> >   example.foo(bar: 1)
> >   example.foo([1]) # raises ArgumentError since expectation A is checked again (?), and Hash.ruby2_keywords_hash? raises an error since an Array is passed in
> >   ```
> >     
> >   Adding a `return false unless parameter.is_a? Hash` should be sufficient.
> 
> OK. I'll take a look at fixing this now. Thanks.

@wasabigeek I've attempted a fix in https://github.com/freerange/mocha/pull/562/commits/95dfe64baf6d1545645e5b3358ba379d74f39f0b. Can you take a look and see if it makes sense to you?

wasabigeek:
It would be nice if we could find a more precise test 🤔 Potentially we could check if that the correct error is returned if the parameter is not a Hash? Although that might be better served by a unit test.

I'm good with this too, maybe we could add a comment explaining the test?

floehopper:
@wasabigeek I'm now leaning towards releasing this PR as part of a **minor** version bump, but initially as a pre-release (e.g. `v1.16.0.alpha`) so we can get a few people to check it's working OK for them, before large numbers of people start installing it.

wasabigeek:
Sounds good 👍

floehopper:
> It would be nice if we could find a more precise test 🤔 

Agreed. I've done this by making [this change](https://github.com/freerange/mocha/commit/eec7200a2cc5fc11a525d0e5156674c7680d36ff) in `main` and then [making use of the now more specific `assert_failed` in the acceptance test](https://github.com/freerange/mocha/pull/562/commits/10d4b24f5184cc86db65e0a0058d1b3a1fab1e59). I'm not entirely happy with it, but I think it's probably an improvement. What do you think...?

I think I'm going to write some unit tests too now.

test/acceptance/keyword_argument_matching_test.rb
```diff
@@ -0,0 +1,177 @@
...
+  if Mocha::RUBY_V27_PLUS
+    def test_should_not_match_non_hash_args_with_keyword_args
+      test_result = run_as_test do
+        mock = mock()
+        mock.expects(:method).with(**{ :key => 1 })
+        Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+          mock.method([2])
+        end
+      end
+      assert_failed(test_result)
```
wasabigeek, line 174: Would we need to test the error message here? In the buggy code, it would still raise an error, but the wrong one (an ArgumentError would be raised by `ruby2_keywords_hash?` instead of the error from mismatched invocation), so I assume the failed result would be returned in both cases (before and after the fix).

floehopper:
@wasabigeek Unit test added in https://github.com/freerange/mocha/pull/562/commits/06663b395eb36a09733e613be30d175ed2a6eae1. Does it make sense to you...?

test/unit/parameter_matchers/positional_or_keyword_hash_test.rb
```diff
@@ -0,0 +1,63 @@
...
+        assert matcher.matches?([{ :key_1 => 1, :key_2 => 2 }, %w[a b]])
+      end
+    end
+
+    def test_should_not_match_non_hash_arg_with_hash_arg_when_strict_keyword_args_is_enabled
+      matcher = { :key_1 => 1, :key_2 => 2 }.to_matcher
+      Mocha::Configuration.override(:strict_keyword_argument_matching => true) do
+        assert !matcher.matches?([%w[a b]])
+      end
+    end
```
wasabigeek, line 33: I think these tests apply without strict keyword arguments as well

wasabigeek (approved):
LGTM, one small comment on the tests 🙏

wasabigeek:
> I've got an idea for how we might be able to generate deprecation warnings which might be a better solution for this.

Nice, let me know if I can help. Does that mean strict keyword matching could potentially be the only behaviour going forward? If so, one thing I hadn't considered was whether we should enforce it for users on Ruby 2.7. Technically, that version still supports loose keyword matching (albeit with deprecation warnings if I recall correctly), so it might be a bit _too opinionated_ of us to force it upon them.

floehopper:
@wasabigeek 

> > I've got an idea for how we might be able to generate deprecation warnings which might be a better solution for this.
> 
> Nice, let me know if I can help.

I'm now a bit less sure about deprecation warnings completely _replacing_ the need for the `with_strict_keywords` idea, but they might be complementary. I'm planning to do some investigation this afternoon. I'll let you know how I get on.

> Does that mean strict keyword matching could potentially be the only behaviour going forward? If so, one thing I hadn't considered was whether we should enforce it for users on Ruby 2.7. Technically, that version still supports loose keyword matching (albeit with deprecation warnings if I recall correctly), so it might be a bit _too opinionated_ of us to force it upon them.

That's a good question. So how about something like this:

### v1.16.0.alpha <= Mocha < v2

* Ruby < v2.7: No configuration option available
* Ruby >= v2.7: Configuration option available, defaulting to `false`; deprecation warnings if option is `false` and a matcher would return a different result if the option were set to `true`.

### Mocha >= v2

* Ruby < v2.7: No configuration option available
* Ruby = v2.7: Configuration option available, defaulting to `true`; no deprecation warnings
* Ruby >= v3: No configuration option available, effectively defaults to `true`

### Questions

I'm assuming it doesn't make sense to be able to set the configuration option to `false` in Ruby v3 - does that sound correct?

Does my suggestion of how the deprecation warnings might work make sense? I think it _might_ highlight some things that wouldn't actually break, but it's the best idea I've come up with so far.

floehopper:
I've opened #563 to illustrate what I was thinking about deprecation warnings.

Regarding the `with_strict_keywords` idea, I've thought about a couple of approaches:

1. If we actually add a new method to `Expectation` like `#with_strict_keywords` then I _think_ we'd have to pass in a `strict_keywords` boolean parameter into the `#to_matcher` method on _all_ the matchers and then pass it into the constructor of `PositionalOrKeywordHash` from `Hash#to_matcher`. Is that what you were imagining?
2. I haven't thought about this in detail, but perhaps we could provide a new API method, e.g. `API#strict_kwargs(hash)` which would wrap the `Hash` in an instance of `PositionalOrKeywordHash` with strict matching enabled. So you could replace something like `mock.expects(:method).with(:key => 42)` with `mock.expects(:method).with(strict_kwargs(:key => 42))`.
3. There is part of me that's wondering whether we definitely need something like `with_strict_keywords`, because it would be possible to have `strict_keyword_argument_matching` set to `false` globally for a test suite, but then use `Configuration.override` to set it to `true` for individual tests as you fix things. I realise it's a bit verbose, but you could probably extract the verbosity into a helper method; in fact we could provide such a helper method as an API method in Mocha...?

@wasabigeek Do you have any thoughts or preferences...?

wasabigeek:
👋
> Does my suggestion of how the deprecation warnings might work make sense?

It looks mostly good to me! I've mixed feelings about defaulting to strict matching for Ruby 2.7 users in Mocha >= v2 though 🤔 On one hand, I think it smoothens the user's upgrade path to Ruby 3 in the future (e.g. if Ruby 2.7 user upgrades to Mocha v2 before upgrading to Ruby 3). On the other, since Ruby 2.7 still allows interchanging hashes and keyword args, we'd be pushing our opinions onto users, even though they can turn strict matching off. I _think_ the former is less pain in the long run, but you may have to deal with some developer ire? 🤣

Either way, we might want to include some upgrading guide when Mocha v2 comes out, especially for Ruby 2.7 users.

> Regarding the with_strict_keywords idea

I do like the idea of Option 3, because a user could stick it in a `before_setup`/`after_teardown` block, which is relatively easier to remove after all fixes are done and strict matching is enabled. Maybe the only assumption is that we'll be keeping the current `with` API (thus Option 1 & 2 would be reverted after a user upgrades), which I think is a sensible assumption.

For our specific use case though, the deprecation warnings work well enough. I spoke with my colleague again, and we can use something like https://github.com/Shopify/deprecation_toolkit to record the warnings and slowly work off the list before the major upgrade.

floehopper:
@wasabigeek 

> > Does my suggestion of how the deprecation warnings might work make sense?
> 
> It looks good to me! I've mixed feelings about defaulting to strict matching for Ruby 2.7 users in Mocha >= v2 though 🤔 On one hand, I think it smoothens the user's upgrade path to Ruby 3 in the future (e.g. if Ruby 2.7 user upgrades to Mocha v2 before upgrading to Ruby 3). On the other, since Ruby 2.7 still allows interchanging hashes and keyword args, we'd be pushing our opinions onto users (but they can turn strict matching off, of course). I _think_ your proposal of defaulting to strict matching is still less pain in the long run, but you may have to deal with some developer ire? 🤣
> 
> Either way, we might want to include some upgrading guide when Mocha v2 comes out, especially for Ruby 2.7 users.

I see what you mean, although I would've thought that since Ruby v2.7 users would have been seeing deprecation warnings and they were electing to do a major version bump of Mocha, that it would be reasonable to expect things to break if they hadn't fixed the deprecation warnings. No?

Anyway, I don't think we have to make a decision about that imminently. I've only just remembered the next release will drop support for Ruby v1.9. Although I (possibly slightly accidentally 😊) dropped support for Ruby v1.8 in a Mocha minor version bump (i.e. v1.15.0), so I'm now leaning more towards including the dropping of support for Ruby v1.9 and the addition of the `strict_keyword_argument_matching` option in a Mocha major version bump (i.e. v2.0.0).

> > Regarding the with_strict_keywords idea
> 
> I do like the idea of Option 3, because a user could stick it in a `before_setup`/`after_teardown` block, which is relatively easier to remove after all fixes are done and strict matching is enabled. Maybe the only assumption is that we'll be keeping the current `with` API (thus a user would revert their changes made in Option 1 & 2 after upgrading successfully), which I think is a sensible assumption.

Hmm. Perhaps you've misunderstood me. I was considering implementing only one of the 3 options, so for option 3 there would only be `Expectation#with`; no alternative like `Expectation#with_strict_keywords`. Does that make sense?

> For our specific use case though, the deprecation warnings work well enough. I spoke with my colleague again, and we can use something like https://github.com/Shopify/deprecation_toolkit to record the warnings (as long as they use `Kernel#warn`) and slowly work off the list before the major upgrade.

Thanks for soliciting that feedback. Does this mean the deprecation warnings would be sufficient on their own, or would you need at least one of the 3 options described above?

Perhaps we can flesh this out more once you've got a pre-release to work with...?

floehopper:
> I spoke with my colleague again, and we can use something like https://github.com/Shopify/deprecation_toolkit to record the warnings (as long as they use `Kernel#warn`) and slowly work off the list before the major upgrade.

Mocha deprecation warnings do use `Kernel#warn`:

https://github.com/freerange/mocha/blob/2e72eb5b23f55142f4eab182817adbf8aaaa35b1/lib/mocha/deprecation.rb#L14

wasabigeek:
> since Ruby v2.7 users would have been seeing deprecation warnings and they were electing to do a major version bump of Mocha, that it would be reasonable to expect things to break if they hadn't fixed the deprecation warnings

I agree, it's reasonable that users should have enough warning of the upcoming change. What I'm unsure of is whether they'd feel like it's unnecessarily forced upon them, since Ruby 2.7 still allows hashes and kwargs to be interchanged. That said, it's still configurable so I'm likely overthinking 😄, and I do think this is the least complicated approach overall. 

How do the deprecation warnings come into play now though, given that the matching is in Mocha v2? Would the "enforcement" come in v3?

> I was considering implementing only one of the 3 options, so for option 3 there would only be Expectation#with; no alternative like Expectation#with_strict_keywords.

This was my understanding as well 👍. I was trying to say that if we picked Option 1 or 2, the user might have to go back and revert a lot of `with_strict_keywords` etc., so I'm all for Option 3!

> Does this mean the deprecation warnings would be sufficient on their own, or would you also need at least one of the 3 options described above?

For us, deprecation warnings would be sufficient on their own, but specifically because most of our codebases are already on Ruby 3 (thus, we'd get deprecation warnings for sure). I think there might be some paths that lead to a rude shock when upgrading Mocha (e.g. a user on Ruby < 2.7 upgrades to Mocha v2, then upgrades to Ruby >= 2.7), not too sure what the ideal approach would be in that scenario (hence the suggestion for an upgrade guide).

floehopper:
> I agree, it's reasonable that users should have enough warning of the upcoming change. What I'm unsure of is whether they'd feel like it's unnecessarily forced upon them, since Ruby 2.7 still allows hashes and kwargs to be interchanged. That said, it's still configurable so I'm likely overthinking 😄, and I do think this is the least complicated approach overall.

Ah, I think I see what you mean now - sorry. Yes, perhaps I should give that a little more thought.

> How do the deprecation warnings come into play now though, given that the matching is in Mocha v2? Would the "enforcement" come in v3 instead?

Yes, I think any enforcement would probably have to come in another major version bump, but I think I need to get over my aversion to major version bumps (!) i.e. it doesn't necessarily mean it would be a long time from now.
 
> This was my understanding as well 👍. I was trying to say that if we picked Option 1 or 2, the user might have to go back and revert a lot of `with_strict_keywords` etc., so I'm all for Option 3!

Ah, yes - we were both thinking the same thing - just got our wires crossed! 😅 

> For us, deprecation warnings would be sufficient on their own, but specifically because most of our codebases are already on Ruby 3 (thus, we'd get deprecation warnings for sure).

That's useful to know.

> I think there might be some paths that lead to a rude shock (e.g. a user on Ruby < 2.7 upgrades to Mocha v2, then upgrades to Ruby >= 2.7), not too sure what the ideal approach would be in that scenario (hence the suggestion for an upgrade guide).

Yes, that makes sense - thanks!

wasabigeek:
To clarify on Ruby 2.7 users, I do think it's to their benefit in the long run for us to give deprecation warnings / default to strict matching. They'd run into then anyway when they upgrade to Ruby 3, so may as well fix it earlier. I'm just not sure if it's going to be a popular opinion! There's still a way out with the config though, so maybe that's enough.

floehopper:
> To clarify on Ruby 2.7 users, I do think it's to their benefit in the long run that we give deprecation warnings / default to strict matching. They'd run into then anyway when they upgrade to Ruby 3, so may as well fix it earlier. I'm just not sure if it's going to be a popular opinion! There's still a way out with the config though, so maybe that's enough.

Yeah, I'm now a bit less convinced about defaulting the config option to true for Ruby v2.7 in a later version of Mocha. I think the deprecation warnings are fine though.

floehopper:
@wasabigeek I've just released this in [Mocha v2.0.0.alpha](https://rubygems.org/gems/mocha/versions/2.0.0.alpha). It would be great if you and your colleagues could try this out on some large repos and let me know if you run into any problems or things that could be made smoother. Thanks again for all your help with this feature!

floehopper:
@wasabigeek I've just released [Mocha v2.0.0.alpha.1](https://rubygems.org/gems/mocha/versions/2.0.0.alpha.1) to improve the deprecation warnings by including the source location where the relevant stub was defined. It would be great if you or your colleagues could try this out and let me know if you run into any problems.

wasabigeek:
Thanks, I'll work on this but may take some time!

Issue #578, opened 2022-10-21
Possible enhancement for responds_with matcher?


vlad-pisanov:
Currently, the matcher `responds_with` takes 2 params: an attribute and a value, like:

```ruby
responds_with(:name, 'Joe')
```

How about introducing a variation of this matcher that takes multiple attribute-value pairs? e.g.

```ruby
responds_with(name: 'Joe', age: 21, salary: 40_000)
```

The matcher would succeed if all attributes match. This would give us a compact way of matching complex objects, and the syntax would be symmetrical to the `has_entries` matcher that we currently have for Hashes:

```ruby
has_entries(name: 'Joe', age: 21, salary: 40_000)
```

floehopper:
@vlad-pisanov Thanks for the suggestion. At first glance it sounds sensible to me. I'll take a proper look at it early next week.

floehopper:
@vlad-pisanov I haven't got much time at the moment, but I managed to find a few minutes to draft [this PR](https://github.com/freerange/mocha/pull/579). Is that roughly what you had in mind?

vlad-pisanov:
@floehopper nice, that's exactly it! I like the idea of using `AllOf.new(...)` under the hood, very semantic

floehopper:
Closed by #579. Sorry it's taken so long!

floehopper:
[Released in v2.2.0](https://github.com/freerange/mocha/blob/main/RELEASE.md#220)

PR #579, opened 2022-10-25
Support multiple methods in responds_with matcher

floehopper:
Addresses #578.

### DIFF
--- a/lib/mocha/parameter_matchers/responds_with.rb
+++ b/lib/mocha/parameter_matchers/responds_with.rb
@@ -1,12 +1,18 @@
 require 'mocha/parameter_matchers/base'
+require 'mocha/parameter_matchers/all_of'
 require 'yaml'
 
 module Mocha
   module ParameterMatchers
-    # Matches any object that responds to +message+ with +result+. To put it another way, it tests the quack, not the duck.
+    # @overload def responds_with(message, result)
+    #   Matches any object that responds to +message+ with +result+. To put it another way, it tests the quack, not the duck.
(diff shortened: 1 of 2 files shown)


lib/mocha/parameter_matchers/responds_with.rb
```diff
@@ -22,8 +28,29 @@ module ParameterMatchers
...
+        raise ArgumentError, 'No arguments. Expecting at least one.'
+      when 1
+        option = options.first
+        raise ArgumentError, 'Argument is not a Hash.' unless option.is_a?(Hash)
+        raise ArgumentError, 'Argument has no entries.' if option.empty?
+
+        messages_vs_results = option
+      when 2
+        messages_vs_results = Hash[*options]
+      else
```
vlad-pisanov, line 49: mayhaps putting the return value logic into the `when` branch itself would make it a bit more clear? something like:

```ruby
when 1
  # ... argument sanity checks ... #
  matchers = option.map { |message, result| RespondsWith.new(message, result) }
  AllOf.new(*matchers)
when 2
  message, result = options
  RespondsWith.new(message, result)
```

either way thanks for adding this 🙌

floehopper: Ah, yes - that's nicer - thanks for the suggestion! 👍

floehopper: Addressed in https://github.com/freerange/mocha/pull/579/commits/4ce03e9ce44e669863832b8417cf6fe58b0f5ceb

PR #582, opened 2022-11-02
ruby2_keywords is runtime dependency; not development dependency

### LINKED ISSUE #581: cannot load such file -- ruby2_keywords (LoadError) on Ruby 2.6
I ran into a LoadError with mocha 2.0.0 and Ruby 2.6.

```
# ruby --version
ruby 2.6.10p210 (2022-04-12 revision 67958) [x86_64-linux-musl]
# irb
irb(main):001:0> require 'mocha/api'
Traceback (most recent call last):
       16: from (irb):1
       15: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:34:in `require'
       14: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:130:in `rescue in require'
       13: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:130:in `require'
       12: from /usr/local/bundle/gems/mocha-2.0.0/lib/mocha/api.rb:3:in `<top (required)>'
       11: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
       10: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        9: from /usr/local/bundle/gems/mocha-2.0.0/lib/mocha/hooks.rb:1:in `<top (required)>'
        8: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        7: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        6: from /usr/local/bundle/gems/mocha-2.0.0/lib/mocha/mockery.rb:2:in `<top (required)>'
        5: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        4: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        3: from /usr/local/bundle/gems/mocha-2.0.0/lib/mocha/mock.rb:1:in `<top (required)>'
        2: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        1: from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
LoadError (cannot load such file -- ruby2_keywords)
```

ruby2_keywords seems to be bundled from Ruby 2.7.
Should it be added to runtime dependency on [mocha.gemspec](https://github.com/freerange/mocha/blob/c5cf3249d9706f3470cbfcfd76e97b4bae87a3d0/mocha.gemspec) rather than [Gemfile](https://github.com/freerange/mocha/blob/c5cf3249d9706f3470cbfcfd76e97b4bae87a3d0/Gemfile#L32)?

floehopper:
And so should be included as such in `mocha.gemspec`; not in the `Gemfile`.

I've specified the '>= 0.0.5' constraint, because v0.0.5 is the version we've been testing with.

I've used a '>=' constraint rather than a twiddle-wakka ('~> 0.0.5'), because that's equivalent to the combination of '>= 0.0.5' and '< 0.1.0' which seems a bit too restrictive. At the moment there's no reason to suppose later versions of `ruby2_keywords` won't work with Mocha and we can always add a constraint later if it turns out there is a problem.

Closes #581.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,5 +28,3 @@ if ENV['MOCHA_GENERATE_DOCS']
   gem 'redcarpet'
   gem 'yard'
 end
-
-gem 'ruby2_keywords', '~> 0.0.5'
--- a/mocha.gemspec
+++ b/mocha.gemspec
@@ -21,4 +21,6 @@ Gem::Specification.new do |s|
   s.homepage = 'https://mocha.jamesmead.org'
   s.require_paths = ['lib']
   s.summary = 'Mocking and stubbing library'
+
+  s.add_runtime_dependency 'ruby2_keywords', '>= 0.0.5'
 end


floehopper:
@mishina2228 I think this addresses your issue. I reproduced the problem locally in a blank project with Ruby v2.6 and this branch seems to fix it, but it would be good to confirm. So, if you get a chance, could you try pointing your `Gemfile` at this version of mocha on the `ruby2-keywords-runtime-dependency` branch and see if fixes the issue?

floehopper:
@wasabigeek Does this look OK to you? In particular does my use of the `'>= 0.0.5'` (vs `'~> 0.0.5`) constraint make sense to you?

wasabigeek:
> In particular does my use of the '>= 0.0.5' (vs '~> 0.0.5) constraint make sense to you?

I personally lean towards ["pessimistic version constraints"](https://guides.rubygems.org/patterns/#pessimistic-version-constraint), especially since it's not a "development dependency". I guess the downside is the dependency will need to be bumped in the future as well (new releases etc).

"~> 0.0.5" should be equivalent to ">= 0.0.5, < 0.1", which should be safer. Though I guess if we strictly follow semver, then any bump in 0.x.x can potentially be breaking (https://semver.org/#spec-item-4).

floehopper:
> I personally lean towards ["pessimistic version constraints"](https://guides.rubygems.org/patterns/#pessimistic-version-constraint), especially since it's not a "development dependency". I guess the downside is the dependency will need to be bumped in the future as well (new releases etc).

I also tend to use pessimistic version constraints, but I normally use them with a minor version (e.g. `~> 1.3` which allows any v1.x from v1.3 onwards but not v2.x or above). In general it seems a bit odd to use it with a *patch* version, since a minor version bump (e.g. to v0.2.0 in the case of `ruby2_keywords`) ought not to introduce any incompatible changes. Do you see what I mean?

> "~> 0.0.5" should be equivalent to ">= 0.0.5, < 0.1", which should be safer. Though I guess if we strictly follow semver, then any bump in 0.x.x can potentially be breaking (https://semver.org/#spec-item-4).

Oh, yes - I forgot the "all bets are off" v0.x thing. Good point!

mishina2228:
@floehopper I confirmed the problem has been fixed.
Thanks for your quick work!

```
$ docker pull ruby:2.6.10-alpine
$ docker run -it --rm ruby:2.6.10-alpine /bin/sh
/ # apk add git
/ # vi Gemfile
```

```ruby
source 'https://rubygems.org'

# gem 'irb' # alpine does not have necessary libraries to install io-console, so I had to use pry.
gem 'pry'
gem 'mocha', git: 'https://github.com/freerange/mocha', branch: 'ruby2-keywords-runtime-dependency'
```

```
/ # bundle
/ # bundle exec pry
[1] pry(main)> require 'mocha/api'
=> true
```

wasabigeek:
> In general it seems a bit odd to use it with a patch version, since a minor version bump (e.g. to v0.2.0 in the case of ruby2_keywords) ought not to introduce any incompatible changes.

Ah yes, ignoring the "all bets are off"-nature of version 0, maybe a more accurate pessimistic versioning would be "< 1"? I'm good to go with your preference though, you definitely have more experience with this than I do 😄

Issue #580, opened 2022-11-01
[Bug] Recent changes to Mock `respond_to?` behaviour causing `NoMethodError`


adrianna-chang-shopify:
Hi!

In trying to upgrade Shopify's primary Rails application from mocha v1.14 to v1.15, a test started to fail unexpectedly. I've set up a repro to mimic what the test was doing (Ruby v3.1.2, but shouldn't matter):

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", "~> 7.0.4"
  gem "sqlite3"
  gem "mocha"
end

require "active_record"
require "minitest/autorun"
require 'mocha/minitest'

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

ActiveRecord::Schema.define do
  create_table :posts, force: true
end

class Post < ActiveRecord::Base; end

class MockingTest < Minitest::Test
  def test_mocking_active_records_flatten
    m1 = mock()
    m1.responds_like(Post.new)

    m2 = mock()
    m2.responds_like(Post.new)

    arr = [m1, m2].flatten
  end
end

=> # Running:

E

Error:
MockingTest#test_mocking_active_records_flatten:
NoMethodError: undefined method `to_ary' for #<Mock:0x2530> which responds like #<Post:0x2558>

        raise NoMethodError, "undefined method `#{symbol}' for #{mocha_inspect} which responds like #{@responder.mocha_inspect}"
        ^^^^^
    /Users/adriannachang/.gem/ruby/3.1.2/gems/mocha-2.0.0/lib/mocha/mock.rb:385:in `check_responder_responds_to'
    /Users/adriannachang/.gem/ruby/3.1.2/gems/mocha-2.0.0/lib/mocha/mock.rb:322:in `handle_method_call'
    /Users/adriannachang/.gem/ruby/3.1.2/gems/mocha-2.0.0/lib/mocha/mock.rb:314:in `method_missing'
    scripts/mocha_experiment.rb:36:in `flatten'
    scripts/mocha_experiment.rb:36:in `test_mocking_active_records_flatten'

rails test scripts/mocha_experiment.rb:29
```

I traced it back to [this change](https://github.com/freerange/mocha/commit/365a7348dbd8e8798584001e261d31a7355c19d8). To sum:
[...]

PR #583, opened 2022-11-04
Fix regression in Mock#responds_like behaviour

floehopper:
The regression was inadvertently introduced in [this commit][1] and indirectly led to [this issue][2].

I've attempted to thoroughly characterise the pre-regression behaviour of `Mock#responds_like` in a new acceptance test, `RespondsLikeTest`, in order to reduce the chances of any such regressions occurring in the future. I've also added another acceptance test, `ArrayFlattenTest`, which reproduced [the issue][2] until the regression was fixed.

Investigating this has made me realise that the behaviour of `Mock#responds_like` in combination with protected & private methods on the responder was not specified in the documentation. I've attempted to reverse engineer the documentation so that it more fully describes the pre-regression behaviour.

Note that the CI build for Ruby v1.9 is using an old version of Minitest which doesn't work with the `RespondsLikeTest` and since support for Ruby v1.9 was dropped in Mocha v2, I don't think it's worth the effort to sort this out so I've just disabled the test for Ruby v1.9.

[1]: https://github.com/freerange/mocha/commit/365a7348dbd8e8798584001e261d31a7355c19d8
[2]: https://github.com/freerange/mocha/issues/580

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 1.15.1
+
+### External changes
+
+* Fix regression in `Mock#responds_like` behaviour - thanks to @adrianna-chang-shopify for reporting (#580,#583,c586a08c)
+
 ## 1.15.0
 
 ### External changes
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -182,11 +182,11 @@ module Mocha
       end
     end
 
-    # Constrains the {Mock} instance so that it can only expect or stub methods to which +responder+ responds. The constraint is only applied at method invocation time.
+    # Constrains the {Mock} instance so that it can only expect or stub methods to which +responder+ responds publicly. The constraint is only applied at method invocation time.
     #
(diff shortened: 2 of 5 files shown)


floehopper:
This is a first stab at fixing the issue highlighted in #580. However, there are still some things to resolve:

* [ ] Does it actually fix #580 ?
* [ ] Should we display a warning when `Mock#respond_to_missing` is called with `include_all` set to `true`?
* [ ] Is there still an issue with flattening an array containing a simple mock object in earlier versions of Ruby due to different `Array#flatten` behaviour?
* [ ] Does it make sense to release this fix as v1.15.1 patch release?
* [ ] Would it then also make sense to port the fix to a v1.16.1 patch release?
* [ ] Should we retain the existing potentially confusing behaviour for protected & private methods?
  * [ ] If no, we'll need to come up with a deprecation strategy and migration path
  * [ ] If yes, it still might be worth trying to further improve the documentation, e.g. extra examples
  * [ ] Is there some crossover with #531 and/or #532 that we need to consider here?

/cc @chrisroos

floehopper:
This fix has been released in v1.15.1, v1.16.1 & v2.0.2. Closing.

Issue #591, opened 2022-11-28
Possible regression in JRuby v9.4.0.0 relating to keyword arguments


floehopper:
This possible regression was highlighted by the scheduled weekly CircleCI build.

I've opened this issue partly to track [this JRuby issue](https://github.com/jruby/jruby/issues/7492) which I have opened and partly to track our own investigations.

* [ ] Revert b8e6d064b8b9bfeb37d6534bbb4e37856dfec5b7 when this fix is actually available in `jruby:latest` on CircleCI

floehopper:
These tests seem to be passing when I install `jruby-dev` via `asdf`. However, the CircleCI build is still failing. I've added a comment on jruby/jruby#7492 to ask about the status of the issue.

floehopper:
jruby/jruby#7492 has now been fixed, but I'm not sure at what point it will be included in the `jruby:latest` on CircleCI. I just tried re-running the failed weekly build and it still failed and it was clear the build was still using the same SHA for JRuby.

floehopper:
It looks as if this will be released in [JRuby v9.4.1.0](https://github.com/jruby/jruby/milestone/111) which is currently at 30% complete.

floehopper:
[This toot](https://ruby.social/@JRuby/109875612330715953) suggests a new release it imminent:

> Looking to do a quick flip JRuby 9.4.2 release very soon. If you have unaddressed issues blocking you, let us know. If you haven't tried updating to 9.4, now's the time to help us find undiscovered problems.

floehopper:
The fix for this was released in v2.0.3.

PR #592, opened 2022-12-02
Fix BacktraceFilter to handle special characters

casperisfine:
We do regular testing of ruby-head, so our gem paths include a `+` which has a special meaning in Regexp.

But either way the regexp capability doesn't seem used, so it's much simpler to just assert the path starts with the lib dir.

cc @floehopper @wasabigeek

### DIFF
--- a/lib/mocha/backtrace_filter.rb
+++ b/lib/mocha/backtrace_filter.rb
@@ -3,11 +3,11 @@ module Mocha
     LIB_DIRECTORY = File.expand_path(File.join(File.dirname(__FILE__), '..')) + File::SEPARATOR
 
     def initialize(lib_directory = LIB_DIRECTORY)
-      @path_pattern = Regexp.new(lib_directory)
+      @lib_directory = lib_directory
     end
 
     def filtered(backtrace)
-      backtrace.reject { |location| @path_pattern.match(File.expand_path(location)) }
+      backtrace.reject { |location| File.expand_path(location).start_with?(@lib_directory) }
     end
   end
 end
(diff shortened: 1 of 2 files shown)


floehopper:
@casperisfine 

Thanks for opening this PR. It makes sense to me. Unfortunately it looks as if [CircleCI is having some issues](https://status.circleci.com/) at the moment, but if/when the build passes I'm happy to merge this. 👍

casperisfine:
I ran the test suite locally and it was green.

Not sure what the problem with CircleCI can be I know very little of it.

casperisfine:
> Could not find a usable config.yml, you may have revoked the CircleCI OAuth app.
Please sign out of CircleCI and log back in with your VCS before triggering a new pipeline.

This sounds like something you have to do?

floehopper:
@casperisfine 

I'm not sure what's going on with CircleCI - it was working fine. As a workaround I've pushed your branch to the `freerange/mocha` repo on GitHub and that [successfully triggered a CircleCI build](https://app.circleci.com/pipelines/github/freerange/mocha/336). There are a few test failures for JRuby, but they relate to an existing regression in JRuby (#591), so I've manually merged the branch into `main`.

Thanks again for taking the time to submit the PR - much appreciated!

floehopper:
@casperisfine:

This has now been released in v2.0.3. Sorry for the delay - my motivation to work on Mocha has been limited recently due to lack of financial support.

casperisfine:
Thanks for the release, and no worries, it's easy enough to run from a branch.

PR #599, opened 2022-12-25
Rubocop: resolve Style/EachWithObject

herwinw:
Ruby 1.9 is the lowest supported version, so this is always available.

But the actual implementation was using a Hash lookup for the keys, but ignored the values. Set is a better datatype for this.

The change has been tested on the command line:
```
~/code/mocha$ bundle exec ruby -e 'require "mocha"; require "mocha/debug"; Mocha::Debug.puts("test")'
~/code/mocha$ MOCHA_OPTIONS="foo,debug,bar" bundle exec ruby -e 'require "mocha"; require "mocha/debug"; Mocha::Debug.puts("test")'
test
~/code/mocha$ 
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,10 +11,6 @@ Style/Alias:
 Style/Documentation:
   Enabled: false
 
-# Enumerable#each_with_object only available since Ruby v1.9
-Style/EachWithObject:
-  Enabled: false
-
 # Kernel#__dir__ has only been available since Ruby v2.0
 Style/ExpandPathArguments:
   Enabled: false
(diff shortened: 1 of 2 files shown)


herwinw:
https://ruby-doc.org/stdlib-1.9.1/libdoc/set/rdoc/Set.html

`Set` should be available since Ruby 1.9.1, not in Ruby 1.9.0

floehopper:
@herwinw Thanks for this - I've gone with an `Array` instead in #600 to avoid having to worry about `Set` not being available in Ruby v1.9.0. I've amended your commit from this PR in #600 so that you get credit as the commit author, but let me know if you're unhappy with the changes I've made and I won't merge it.

herwinw:
I'm totally fine with this change. Given the expected amount of values in the envvar, using an Array is probably faster too.

PR #600, opened 2022-12-26
Simplify storage of MOCHA_OPTIONS

floehopper:
This has only ever supported `debug`, `verbose` & `skip_integration` options. However, it currently only supports `debug`.

Storing these options in a `Hash` was always a bit over the top and since it was the only reason we needed to disabled the `Style/EachWithObject` Rubocop rule, it seems more sensible to use an `Array` instead.

I don't think that the "uniqueness" of the options in the comma-separated list was ever a motivation for storing them in a `Hash` and so an `Array` (rather than a `Set`) seems good enough.

Supersedes #599.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,10 +11,6 @@ Style/Alias:
 Style/Documentation:
   Enabled: false
 
-# Enumerable#each_with_object only available since Ruby v1.9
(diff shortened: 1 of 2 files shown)


floehopper:
This was released in v2.0.3.

PR #603, opened 2022-12-28
Rubocop: enable Style/FormatStringToken cop

herwinw:
Since support for Ruby 1.8 has been dropped.


floehopper:
This was released in v2.0.3.

PR #604, opened 2022-12-31
Un-pin JRuby in CircleCI builds

floehopper:
Revert the change in b8e6d064b8b9bfeb37d6534bbb4e37856dfec5b7 so CircleCI builds pick up the latest JRuby version again.

This is waiting on the JRuby fix referenced in #591 which is going to be released in JRuby v9.4.1.0. The CircleCI build for this PR will fail until the fix is released.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
       - run: ruby --version
       - when:
           condition:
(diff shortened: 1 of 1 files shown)


floehopper:
This was addressed in 4f5bb2f090df10164ba9502ebd615f1d241cc5b5. Closing.

PR #612, opened 2023-06-03
Remove Google Analytics tracking code

floehopper:
I never look at the data from this and in any case I don't ask permission to track people on the website, so it's long overdue for removal.


floehopper:
This was released in v2.0.3.

Issue #614, opened 2023-07-26
Mocha is incompatible with minitest v5.19 and later


kyrofa:
[Minitest has removed](https://github.com/minitest/minitest/blob/master/History.rdoc) the ancient MiniTest compatibility layer that Mocha has been relying on unless the user defines the `MT_COMPAT` environment variable. See [the commit](https://github.com/minitest/minitest/commit/a2c6c18570f6f0a1bf6af70fe3b6d9599a13fdd6) for more details.

Seems that Mocha should move to using `Minitest` instead?

floehopper:
@kyrofa:

Thanks for the heads up, although I'm a bit confused about what's broken. I just manually ran [the weekly build](https://app.circleci.com/pipelines/github/freerange/mocha/415/workflows/121f5ace-9133-4d22-a073-85a69cde57dd) which tries to pick up on incompatibilities like this, but nothing failed even though [the job using the latest Minitest gem seems to have used v5.19.0](https://app.circleci.com/pipelines/github/freerange/mocha/415/workflows/121f5ace-9133-4d22-a073-85a69cde57dd/jobs/5849) which is the version you mention.

If you understand more about exactly what the problem is, it would be great if you can save me some time and tell me what's actually broken and ideally what I need to change!

floehopper:
And do you know if there was a Minitest deprecation warning for this? If so, I missed it!

manewitz:
here's a backtrace if that's helpful:

```
/Users/mikemanewitz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/mocha-2.0.4/lib/mocha/integration/mini_test/adapter.rb:26:in `included': uninitialized constant MiniTest (NameError)

          Mocha::ExpectationErrorFactory.exception_class = ::MiniTest::Assertion
                                                                     ^^^^^^^^^^^
Did you mean?  Minitest
	from /Users/mikemanewitz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/mocha-2.0.4/lib/mocha/integration/mini_test.rb:21:in `include'
	from /Users/mikemanewitz/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/mocha-2.0.4/lib/mocha/integration/mini_test.rb:21:in `activate'
[...]

PR #615, opened 2023-07-26
WIP: First stab at addressing #614

floehopper:
TODO:

- [ ] Ideally I'd like to add a failing test to force me to make this change
- [ ] Check that other occurrences of MiniTest aren't breaking things

### DIFF
--- a/lib/mocha/integration/mini_test/adapter.rb
+++ b/lib/mocha/integration/mini_test/adapter.rb
@@ -23,7 +23,7 @@ module Mocha
 
         # @private
         def self.included(_mod)
-          Mocha::ExpectationErrorFactory.exception_class = ::MiniTest::Assertion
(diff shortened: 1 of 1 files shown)


Maimer:
`minitest/unit` is no longer being required by `minitest` based on this commit: https://github.com/minitest/minitest/commit/a2c6c18570f6f0a1bf6af70fe3b6d9599a13fdd6

However, in the `test_helper` it is being required here: https://github.com/freerange/mocha/blob/main/test/test_helper.rb#L20

If you remove that line from the `test_helper` (which I believe is the correct thing to do) then you will see the failure due to the issue you are fixing in this PR.

floehopper:
@Maimer Ah, thank you - I haven't had much time to look at this closely, so that's really helpful! 🙏

PR #624, opened 2023-11-11
Avoid rubocop directive ending up in YARD docs

floehopper:
It turns out that YARD displays a special section "Dynamic Method Handling" for classes that implement `method_missing` even though this method is tagged as private. The rubocop directive was getting caught up in the YARD docs.


floehopper:
This is now working correctly:

Screenshot 2023-11-11 at 14 29 28

PR #644, opened 2024-04-10
Include the original test name in expired stub error messages

floehopper:
Based on #641 and thanks to @casperisfine.

It is currently very hard to track down such leaked stubs, when they happen, as they might be created in one test and cause an exception dozens of tests later.

By including the name of the test that invalidated them, it makes the debugging process a bit easier.

Note, I made `setup` also accept the current test instance purely for consistency with `teardown` but it's not strictly necessary.


floehopper:
[Released in v2.2.0](https://github.com/freerange/mocha/blob/main/RELEASE.md#220)

PR #648, opened 2024-05-16
Fix nested parameter matching for keyword arguments

### LINKED ISSUE #647: Parameter matching fails on keyword arguments
Pretty basic but surprising bug that I ran into just now, where parameter matching on `expect(...).with(...)` doesn't work properly with keyword arguments.

```ruby
        class Foo
          def self.bar(a:, b:, c:)
            a + b + c
          end
        end

        def test_mocha_mocking_kw_args
          mock = mock("foo")
          mock.expects(:bar).with(a: 1, b: 2, c: is_a(Integer)).returns(6)

          mock.bar(a: 1, b: 2, c: 3)
        end
```

This results in the following error:

```
unexpected invocation: #<Mock:foo>.bar(:a => 1, :b => 2, :c => 3)
unsatisfied expectations:
- expected exactly once, invoked never: #<Mock:foo>.bar(:a => 1, :b => 2, :c => is_a(Integer))
```

The positional argument equivalent works just fine:

```ruby
        class Foo
          def self.bar(a, b, c)
            a + b + c
          end
        end

        def test_mocha_mocking_positional_args
          mock = mock("foo")
          mock.expects(:bar).with(1, 2, is_a(Integer)).returns(6)

          mock.bar(1, 2, 3)
        end
```

Mocha: v2.2.0
Ruby: 3.3.1

floehopper:
Previously, parameter matching for keyword arguments only worked when the matchers specified the exact value of each argument. This change means that we make use of the `HasEntries` matcher to ensure that any matchers for each of the keyword arguments are also taken into account.

Fixes #647.

### DIFF
--- a/lib/mocha/parameter_matchers/positional_or_keyword_hash.rb
+++ b/lib/mocha/parameter_matchers/positional_or_keyword_hash.rb
@@ -13,7 +13,8 @@ module Mocha
 
       def matches?(available_parameters)
         parameter, is_last_parameter = extract_parameter(available_parameters)
-        return false unless parameter == @value
+
+        return false unless HasEntries.new(@value).matches?([parameter])
 
         if is_last_parameter && !same_type_of_hash?(parameter, @value)
           return false if Mocha.configuration.strict_keyword_argument_matching?
--- a/test/acceptance/keyword_argument_matching_test.rb
+++ b/test/acceptance/keyword_argument_matching_test.rb
@@ -189,6 +189,15 @@ class KeywordArgumentMatchingTest < Mocha::TestCase
     assert_passed(test_result)
   end
 
+  def test_should_match_keyword_args_with_nested_matcher
+    test_result = run_as_test do
+      mock = mock()
+      mock.expects(:method).with(key: is_a(Integer))
+      mock.method(key: 42)
+    end
+    assert_passed(test_result)
+  end
+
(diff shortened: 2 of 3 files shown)


floehopper:
@CodingAnarchy Are you able to confirm that this fixes the problem in #647?

CodingAnarchy:
@floehopper I tested with this branch in our tests, and it works great! Thanks for the quick response!

floehopper:
> @floehopper I tested with this branch in our tests, and it works great! Thanks for the quick response!

Great - thanks for letting me know - I'll try to get this merged and released as soon as I can.

ElvinEfendi:
This is a breaking change, isn't it? If so, did we mean it to be?

The test below fails with mocha `2.2.0` but succeeds in `2.4.0`.

```ruby
require "bundler/setup"
require "minitest/autorun"
require "mocha/minitest"
require "mocha/mock"

  def setup
    @mock = mock("foo")
  end

  def test_mocha_mocking_kw_args_with_has_entry
    @mock.expects(:bar).with(has_entry(:a, k: "v"))

    @mock.bar(a: { k: "v", extra: "extra" })
  end
end
```

One can argue the new behaviour is even buggy, because if I want the above test to pass I should configure my expectation as `@mock.expects(:bar).with(has_entry(:a, has_entry(k: "v")))`.

PR #652, opened 2024-06-16
Improve rendering of keyword arguments

floehopper:
Using the `=>` for them was a bit confusing.

The string representation of the keyword hash `{a: true, b: 'c'}` is now 'a: true, b: "c"' instead of ':a => true, :b => "c"'.

One reason (perhaps the only reason?) we didn't make this change when we introduced keyword argument matching was the theoretical edge case where a keyword-style hash could have a non-symbol key.

However, the changes in this commit handle that scenario while still rendering a less-confusing string representation of a keyword hash in the most common scenario where the keys are all symbols.

I've taken @casoerisfine's changes from #649 and added some unit test coverage and expanded the commit note.

### DIFF
--- a/lib/mocha/inspect.rb
+++ b/lib/mocha/inspect.rb
@@ -19,8 +19,19 @@ module Mocha
 
     module HashMethods
       def mocha_inspect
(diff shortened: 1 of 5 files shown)


floehopper:
Released in [v2.4.0](https://rubygems.org/gems/mocha/versions/2.4.0).

Issue #662, opened 2024-07-22
Regression after upgrading to `2.4.3`


vlad-pisanov:
After upgrading to `2.4.3`, a previously-passing test started failing with a mocha exception. It seems to have to do with mocking two expectations on the same method with two different signatures:

Repro code:
```ruby
class C
   def self.foo(bar: 0); end
end

# test:
C.expects(:foo).with(bar: 42)
C.expects(:foo)
      
C.foo(bar: 42)
C.foo
```

Exception:
```
NoMethodError: undefined method `length' for nil
```

Stack trace:
```
mocha-2.4.3/lib/mocha/parameter_matchers/has_entries.rb:41:in `matches?'
mocha-2.4.3/lib/mocha/parameter_matchers/positional_or_keyword_hash.rb:18:in `matches?'
mocha-2.4.3/lib/mocha/parameters_matcher.rb:21:in `block in parameters_match?'
mocha-2.4.3/lib/mocha/parameters_matcher.rb:21:in `all?'
mocha-2.4.3/lib/mocha/parameters_matcher.rb:21:in `parameters_match?'
mocha-2.4.3/lib/mocha/parameters_matcher.rb:16:in `match?'
mocha-2.4.3/lib/mocha/expectation.rb:647:in `match?'
```

floehopper:
Thanks for reporting - I'll take a look in the morning

floehopper:
@vlad-pisanov I think I've reproduced the problem, although I had to change your code slightly:

```ruby
require 'bundler/setup'
require 'minitest/autorun'
require 'mocha/minitest'

class C
  def self.foo(bar: 0); end
end

class FooTest < Minitest::Test
  def test_foo
    C.expects(:foo).with(bar: 42)
    C.expects(:foo)

    C.foo
    C.foo(bar: 42)
  end
end
```

i.e. I had to reverse the order of the method invocations.

Does that make sense?

vlad-pisanov:
I get the same `NoMethodError: undefined method `length' for nil` exception with both
```ruby
C.foo
C.foo(bar: 42)
```
and
```ruby
C.foo(bar: 42)
C.foo
```

floehopper:
@vlad-pisanov Ah, yes - my mistake - I was working from a simplified version. Please ignore me!

floehopper:
[...]

PR #664, opened 2024-07-24
Fix regression when stubbed method expects Hash but receives ActionController::Parameters object

floehopper:
This was broken in [this previous commit][1] which was part of https://github.com/freerange/mocha/pull/660 and which was released in v2.4.3.

Previously the only requirement we put on `Hash`-like arguments was that they respond to `#keys` & `#[]` methods. However, the change in [that commit][1] we introduced an implicit assumption that they also respond to a `#length` method which is caused the problem described [here][2].

However, [`ActionController::Parameters`][3] is commonly used in Rails apps and it does *not* respond to `#length`. So it seems better to add a guard condition to ensure the object responds to `#keys` and use that to determine the length.

Note that the `HasEntry`, `HasKey` and `HasKeys` matchers already work this way and so this feels like it makes `HasEntries` more consistent.

[1]: https://github.com/freerange/mocha/commit/5e6a07b2710dac76e9346def561ca0d44765bf86
[2]: https://github.com/freerange/mocha/issues/662#issuecomment-2246105530
[3]: https://api.rubyonrails.org/classes/ActionController/Parameters.html

### DIFF
--- a/lib/mocha/parameter_matchers/has_entries.rb
+++ b/lib/mocha/parameter_matchers/has_entries.rb
@@ -39,7 +39,8 @@ module Mocha
       def matches?(available_parameters)
         parameter = available_parameters.shift
         return false unless parameter
-        return false if @exact && @entries.length != parameter.length
+        return false unless parameter.respond_to?(:keys)
+        return false if @exact && @entries.length != parameter.keys.length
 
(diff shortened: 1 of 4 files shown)


evgeni (approved):

floehopper:
Released in [v2.4.5](https://rubygems.org/gems/mocha/versions/2.4.5).

PR #669, opened 2024-07-29
Add frozen_string_literal to mockery

radville:
### Description
This adds the magic comment for making string literals frozen in `Mockery`. This PR also purposely allows the string in `#mocha_inspect` to remain unfrozen with by duplicating it since that behavior relies on an unfrozen string. I couldn't use `+''` because that method isn't supported in earlier versions of Ruby.

### Context
There is a deprecation warning for chilled strings merged recently in Ruby: https://github.com/ruby/ruby/commit/12be40ae6be78ac41e8e3f3c313cc6f63e7fa6c4

With that change we started to see a chilled string warning for the string in `#mocha_inspect`.

### DIFF
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'mocha/central'
 require 'mocha/mock'
 require 'mocha/names'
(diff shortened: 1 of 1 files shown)


floehopper:
@radville Many thanks for the nudge to sort this out. I decided to solve it a slightly different way in 55dd0fca85a693d85b95e3fdbf07750988f8e48a and to group it with some other Ruby v3.4-related fixes in #672.

Issue #676, opened 2024-10-17
Improve docs for keyword argument regression introduced in v2.3.0 and corresponding fixes


floehopper:
* Retrospectively add a warning about the regressions & fixes to the release notes for the affected versions
* Add something to [the known issues](https://mocha.jamesmead.org/index.html#known-issues) in the README

PR #677, opened 2024-10-23
Add warnings about regressions to release notes

floehopper:
And move most of the "known issues" from the README into the release notes where they are more likely to be useful.

See #676.


floehopper:
Released in v2.5.0.

Issue #678, opened 2024-11-06
`expects.never` doesn't handle `stubs` methods correctly


ducmtran:
We found that `expects.never` doesn't fail a test when the method is stubbed with `stubs`. Reproduce example below, with the `test_foo_never_called` case. I don't think this is intended?

```ruby
require 'rubygems'
gem 'mocha'
require 'minitest/autorun'
require 'mocha/minitest'

class TestMeme < Minitest::Test
  class Klass
    def self.foo; end
  end

  def test_foo_never_called # should fail, but doesn't
    Klass.stubs(:foo)
    Klass.expects(:foo).never
    Klass.foo
  end

  def test_foo_called_once # correctly passes
    Klass.stubs(:foo)
    Klass.expects(:foo).once
    Klass.foo
  end

  def test_no_stubs # correctly fails
    Klass.expects(:foo).never
    Klass.foo
  end
end
```

output
```
➜ sandbox ruby mocha_stub.rb
Run options: --seed 21862

# Running:

F

Failure:
TestMeme#test_no_stubs [mocha_stub.rb:25]:
unexpected invocation: TestMeme::Klass.foo()
unsatisfied expectations:
- expected never, invoked once: TestMeme::Klass.foo(any_parameters)



rails test mocha_stub.rb:23

..

Finished in 0.000907s, 3307.6069 runs/s, 2205.0713 assertions/s.
3 runs, 2 assertions, 1 failures, 0 errors, 0 skips
```
cc @zenspider

floehopper:
This is working as expected and is effectively a duplicate of #490, #131 & #44. However, I recognize that the behaviour is confusing and, having revisited my responses to those issues, I'm now more sympathetic to trying to come up with a fix. 

Previously I'd hoped that fixing #173 would solve this problem, but I now recognize that I'm unlikely to implement that any time soon.

So to that end I've spiked on a fix in #679 which would generate this extra test failure for your example:

```
  1) Failure:
TestMeme#test_foo_never_called [foo.rb:11]:
not all expectations were satisfied
unsatisfied expectations:
- expected never, invoked once: TestMeme::Klass.foo(any_parameters)
satisfied expectations:
- allowed any number of times, invoked never: TestMeme::Klass.foo(any_parameters)
```

**Is that the kind of failure you are expecting?**

Now I just need to find the time to:
1. Work through the test failures in the spike to see if this fix causes collateral damage elsewhere
2. Work out how to roll this out, i.e. deprecation warnings, etc

[...]

Issue #490, opened 2020-09-15
Unexpected behaviour of `expects.never` on `stub`-ed method


BrentWheeldon:
I'm working on a test that is passing even when it shouldn't. I've simplified it down to a generic example:

```ruby
setup do
  @mock = Mocha::Mock.new(nil)
  @mock.stubs(:a_method)
end

def test_a_method_never_called
  @mock.expects(:a_method).never

  @mock.a_method
end
```

That test passes, which I've found a bit unexpected. We use the `setup` because there are many test cases and most of them just need a simple stub, but in some we want to assert on how many times the method is called.

If I change my test to:

```ruby
setup do
  @mock = Mocha::Mock.new(nil)
  @mock.stubs(:a_method)
end

def test_a_method_never_called
  @mock.unstub(:a_method)
  @mock.expects(:a_method).never

  @mock.a_method
end
```

It fails as expected... But that feels like a bit of a hack?

I wrote a failing test case for [`MockTest`](https://github.com/freerange/mocha/blob/main/test/unit/mock_test.rb) that demonstrates the issue:

```ruby
  def test_expecting_stubbed_method_never_fails
    mock = build_mock
    mock.stubs(:a_method)
    mock.expects(:a_method).never

    mock.a_method

    assert !mock.__verified__?
  end
```

It's entirely possible I'm missing something here so any advice on alternative approaches would be appreciated! (Also, if this is a bug I'm happy to PR a fix if there's guidance on how this might be fixed!)

floehopper:
Hi @BrentWheeldon. Thanks for your question. I'm sorry for the slow reply - really busy with work right now and about to go on holiday.

I haven't looked at your examples in detail, but at first glance I think your scenario is somewhat similar to #171. Since that issue was reported, I've made some changes to the documentation:

From [here](https://mocha.jamesmead.org/#Expectation_matching___invocation_order):

> Stubs and expectations are basically the same thing. A stub is just an expectation of zero or more invocations. The `Expectation#stubs` method is syntactic sugar to make the intent of the test more explicit.
>
> When a method is invoked on a mock object, the mock object searches through its expectations from newest to oldest to find one that matches the invocation. After the invocation, the matching expectation might stop matching further invocations.

And from [here](https://mocha.jamesmead.org/Mocha/Mock.html):

> However, there are some possible “gotchas” caused by this scheme:
> 
> * if you create an expectation and then a stub for the same method, the stub will always override the expectation and the expectation will never be met.
>
[...]

Issue #131, opened 2013-02-01
Can't override stubs


viking:
Consider:

```
require 'rubygems'
gem 'test-unit'
require 'test/unit'
require 'mocha/setup'

class TestFoo < Test::Unit::TestCase
  def setup
    @foo = stub('foo', :bar => 'bar')
  end

  test "foo" do
    @foo.expects(:bar).never
    @foo.bar
  end
end
```

This test passes. Is this intended? (I'm using Ruby 1.9.3, Mocha 0.13.2, and TestUnit  2.5.4)

floehopper:
A fix for this was released in v2.7.0 - better late than never?!

PR #679, opened 2024-11-07
Fail fast if invocation matches never expectation 

floehopper:
#### TODO
* [x] Come up with a version of this that emits a deprecation warning rather than actually changing the behaviour.
* [x] Ensure all the relevant documentation is updated, e.g. method dispatch in README, class docs for `Mock`, and docs for any relevant methods.
* [x] Update commit note(s) to reference being consistent with the behaviour described in the following test which was added in https://github.com/freerange/mocha/commit/d3583772c3fd878a567a0b04750d98ff7d079358

https://github.com/freerange/mocha/blob/f899c034a6f1472788936f2a889ae8d5c323f1de/test/acceptance/expected_invocation_count_test.rb#L202-L214

Previously when an invocation matched an expectation which did not allow invocations (i.e. `Expectation#never` had been called on it), but the invocation also matched another expectation which did allow invocations, then the test would not fail with an unexpected invocation error.

This was happening because neither the `if` condition was `true`, because the "never" expectation was not returned by `ExpectationList#match_allowing_invocation`, but the other expectation allowing expectations was returned. Thus `Expectation#invoke` was called on the latter and `Mock#raise_unexpected_invocation_error` was not called.

This behaviour was confusing and had led to a number of issues being raised over the years: https://github.com/freerange/mocha/issues/44, https://github.com/freerange/mocha/issues/131, https://github.com/freerange/mocha/issues/490 & most recently https://github.com/freerange/mocha/issues/678. Previously I had thought this was a symptom of the JMock v1 dispatch behaviour (which _might_ still be true) and thus would be best addressed by adopting the JMock v2 dispatch behaviour (https://github.com/freerange/mocha/issues/173). However, having considered this specific scenario involving a "never" expectation more carefully, I've decided to try to fix it with the changes in this PR.

Now a test like this will fail with an unexpected invocation error:

    mock = mock('mock')
    mock.stubs(:method)
    mock.expects(:method).never
    mock.method

    unexpected invocation: #<Mock:mock>.method()
    unsatisfied expectations:
    - expected never, invoked once: #<Mock:mock>.method(any_parameters)
    satisfied expectations:
    - allowed any number of times, invoked never: #<Mock:mock>.method(any_parameters)

Closes https://github.com/freerange/mocha/issues/678. Also addresses https://github.com/freerange/mocha/issues/490, https://github.com/freerange/mocha/issues/131 & https://github.com/freerange/mocha/issues/44.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Maybe, but probably not. Partial mocking changes the state of objects in the `Ob
 
 Stubs and expectations are basically the same thing. A stub is just an expectation of zero or more invocations. The `Expectation#stubs` method is syntactic sugar to make the intent of the test more explicit.
 
-When a method is invoked on a mock object, the mock object searches through its expectations from newest to oldest to find one that matches the invocation. After the invocation, the matching expectation might stop matching further invocations.
+When a method is invoked on a mock object, the mock object searches through its expectations from newest to oldest to find one that matches the invocation. After the invocation, the matching expectation might stop matching further invocations. If the expectation that matches the invocation has a cardinality of "never", then an unexpected invocation error is reported.
 
 See the [documentation](https://mocha.jamesmead.org/Mocha/Mock.html) for `Mocha::Mock` for further details.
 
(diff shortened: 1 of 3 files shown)


floehopper:
Released in v2.7.0

PR #681, opened 2024-11-09
Expectation with never cardinality should display deprecation warning

floehopper:
Currently when an invocation matches an expectation which does not allow invocations (i.e. `Expectation#never` has been called on it), but the invocation also matches another expectation which does allow invocations, then the test does not fail with an unexpected invocation error.

In https://github.com/freerange/mocha/pull/679 I'm planning to change this behaviour so the test fails fast with an unexpected invocation error. This PR means we display a deprecation warning instead to give users a chance to update their code before the actual change is made.

### DIFF
--- a/lib/mocha/cardinality.rb
+++ b/lib/mocha/cardinality.rb
@@ -34,6 +34,10 @@ module Mocha
       @invocations.size < maximum
     end
 
+    def invocations_never_allowed?
+      maximum.zero?
+    end
+
     def satisfied?
       @invocations.size >= required
     end
--- a/lib/mocha/expectation.rb
+++ b/lib/mocha/expectation.rb
@@ -654,6 +654,11 @@ module Mocha
     end
 
     # @private
+    def invocations_never_allowed?
+      @cardinality.invocations_never_allowed?
+    end
+
+    # @private
     def satisfied?
       @cardinality.satisfied?
     end
--- a/lib/mocha/expectation_list.rb
+++ b/lib/mocha/expectation_list.rb
@@ -25,6 +25,10 @@ module Mocha
       matching_expectations(invocation).detect(&:invocations_allowed?)
     end
 
+    def match_never_allowing_invocation(invocation)
+      matching_expectations(invocation).detect(&:invocations_never_allowed?)
+    end
+
     def verified?(assertion_counter = nil)
       @expectations.all? { |expectation| expectation.verified?(assertion_counter) }
     end
--- a/lib/mocha/expectation_list.rb
+++ b/lib/mocha/expectation_list.rb
@@ -53,8 +53,6 @@ module Mocha
       self.class.new(to_a + other.to_a)
     end
 
-    private
-
     def matching_expectations(invocation, ignoring_order: false)
       @expectations.select { |e| e.match?(invocation, ignoring_order: ignoring_order) }
     end
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -323,17 +323,20 @@ module Mocha
       check_responder_responds_to(symbol)
       invocation = Invocation.new(self, symbol, arguments, block)
 
-      matching_expectation_allowing_invocation = all_expectations.match_allowing_invocation(invocation)
-      matching_expectation_never_allowing_invocation = all_expectations.match_never_allowing_invocation(invocation)
-      matching_expectation_ignoring_order = all_expectations.match(invocation, ignoring_order: true)
+      matching_expectations = all_expectations.matching_expectations(invocation)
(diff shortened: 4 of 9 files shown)


ducmtran:
we found several failing tests, here's one
```
  def test_stub_with
    Klass.stubs(:foo).with { |_| p 'inside stub' }
    Klass.foo
  end
```
on mocha main branch, it only produces one print statement 
```
# Running:

"inside stub"
"inside stub"
"inside stub"
.
```

i couldn't find the time to debug all the failing cases so there might be other issues - we are pretty busy with other work right now so responses might be delayed, but thanks a lot for diving into this!

cc @zenspider

floehopper:
@ducmtran 

Thanks for running your test suite against the branch with the deprecation warnings. I assume that the "failure" in the the example test is because in reality you have tests _relying_ on `Expectation#with` only being called once. Is that correct? i.e. they will fail if it's called multiple times.

```ruby
 def test_stub_with
    Klass.stubs(:foo).with { |_| p 'inside stub' }
    Klass.foo
  end
```

If so, while this was never [documented behaviour](https://mocha.jamesmead.org/Mocha/Expectation.html#with-instance_method) to be relied upon, I've [changed the code](https://github.com/freerange/mocha/pull/681/commits/065fe3d832e5d98b6bab58601d07a4cb8eb1da81) to avoid multiple invocations of the matcher block. 😓 

Could you try running your test suite against the updated branch when you have time? 🙏

ducmtran:
```
  def test_mocha_never_warning
    Klass.stubs(:foo)
    Klass.expects(:foo).never
    Klass.foo
  end
```
hi, i tried the latest commit again - the above is the original issue we reported and the latest commit is failing it instead of giving a deprecation warning 😅

floehopper:
@ducmtran

Hmm. That's very odd. I have an acceptance test which is pretty much exactly that test and it is passing in the branch. To double-check, I just ran the test below and got the expected output, i.e. the test passed but a deprecation warning was reported.

Can you double-check that you were using the correct branch? i.e. `expectation-with-never-cardinality-should-display-deprecation-warning` (not `expectation-with-never-cardinality-should-fail-fast-if-invoked`) and that it includes the latest commit, i.e. 73c4ea6ef5def4f9203e7a0d6f3a460849d511bb.

```ruby
# klass_test.rb
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "minitest"
  gem "mocha", git: "https://github.com/freerange/mocha.git", branch: "expectation-with-never-cardinality-should-display-deprecation-warning"
end

require "minitest/autorun"
require "mocha/minitest"

class TestKlass < Minitest::Test
  class Klass
    def self.foo; end
  end

  def test_foo_never_called
    Klass.stubs(:foo)
    Klass.expects(:foo).never
    Klass.foo
  end
end
```

```
> ruby klass_test.rb
Run options: --seed 22994

# Running:

Mocha deprecation warning at klass_test.rb:20:in `test_foo_never_called': The expectation defined at klass_test.rb:19:in `test_foo_never_called' does not allow invocations. However, TestKlass::Klass.foo() was invoked. This invocation will cause the test to fail fast in a future version of Mocha.
.

Finished in 0.001337s, 747.9431 runs/s, 747.9431 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

floehopper:
@ducmtran Have you had a chance to look at this? Otherwise I'll probably just release it!

ducmtran:
@floehopper apologies for the late response. Yes Github UI for "referenced this commit" confused me and i accidentally used the commit from #679 🤦 . With the correct commit, everything passed for us, with warning printed locally (we have an issue with our CI setup not printing the warning but i doubt it's relevant to this PR). Thanks!

floehopper:
> @floehopper apologies for the late response. Yes Github UI for "referenced this commit" confused me and i accidentally used the commit from #679 🤦 . With the correct commit, everything passed for us, with warning printed locally (we have an issue with our CI setup not printing the warning but i doubt it's relevant to this PR). Thanks!

Great - thanks for letting me know. 👍

floehopper:
I've tweaked the text of the deprecation warning slightly, tweaked the name of the acceptance test, and expanded the main commit note to explain that the new behaviour will bring the behaviour into line with what is already the case for mocks where `Mock#stub_everything` has been called as per the acceptance test below that was introduced in [this commit][1].

https://github.com/freerange/mocha/blob/f2fa99197f35e2d2ce9554db63f6964057e29ce0/test/acceptance/expected_invocation_count_test.rb#L202-L214

[1]: https://github.com/freerange/mocha/commit/d3583772c3fd878a567a0b04750d98ff7d079358

PR #686, opened 2024-11-27
Fix never expectation deprecation warning logic

floehopper:
In #681 I added logic to display a deprecation warning about a change I planned to make in #679. However, it turns out the logic in both was faulty as demonstrated in [this example][1] where the deprecation warning was incorrectly displayed for the 2nd call to `Foo.create_if`:

```ruby
    class Foo
      def self.create_if(condition)
        new if condition
      end
    end

    test "it creates only if condition is true" do
      Foo.expects(:new).never
      Foo.create_if(false)

      Foo.expects(:new).once
      Foo.create_if(true)
    end
```

This commit adds a new acceptance test to cover an example where the logic was incorrect and updates the logic in `Mock#handle_method_call` to fix the logic so this new test passes and none of the existing tests fail. Unfortunately the implementation is now rather complicated and hard to follow, but I think it will do for now. I have a couple of ideas for ways to simplify it, but I don't have time to work on that now.

I have [updated](https://github.com/freerange/mocha/pull/679/commits/29a8e085407fd50892be5c7f6e2768f019e6d47b) the equivalent logic in #679 to match the new logic implemented here.

[1]: https://github.com/freerange/mocha/pull/679#issuecomment-2503272428

### DIFF
--- a/lib/mocha/mock.rb
+++ b/lib/mocha/mock.rb
@@ -318,25 +318,31 @@ module Mocha
     ruby2_keywords(:method_missing)
 
     # @private
-    def handle_method_call(symbol, arguments, block)
+    def handle_method_call(symbol, arguments, block) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       check_expiry
       check_responder_responds_to(symbol)
       invocation = Invocation.new(self, symbol, arguments, block)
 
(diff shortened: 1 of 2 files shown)


floehopper:
Released in [v2.6.1](https://rubygems.org/gems/mocha/versions/2.6.1).

PR #688, opened 2024-12-06
Support single quote instead of backtick for Ruby 3.4

voxik:
Use a single quote instead of a backtick as an opening quote. [[Feature #16495](https://bugs.ruby-lang.org/]issues/16495)]


floehopper:
@voxik Many thanks for opening this PR. I've made a couple of tweaks and opened #689 to incorporate them. There's a JRuby issue which is currently blocking this from being merged which I have referenced in #689. I'm going to close this PR in favour of #689.

voxik:
Many thanks for taking it further 👍

Issue #628, opened 2023-11-12
Drop support for Ruby v2.1


floehopper:
This version has been [EOL since 31 Mar 2017](https://www.ruby-lang.org/en/downloads/branches/), i.e. nearly 8 years ago.

PR #694, opened 2024-12-08
Deprecate `Configuration#stubbing_method_on_nil=`

### LINKED ISSUE #693: Deprecate `Mocha::Configuration#stubbing_method_on_nil=`
From [the docs](https://mocha.jamesmead.org/Mocha/Configuration.html#stubbing_method_on_nil=-instance_method):

> This option only works for Ruby < v2.2.0. In later versions of Ruby `nil` is frozen and so a [StubbingError](https://mocha.jamesmead.org/Mocha/StubbingError.html) will be raised if you attempt to stub a method on `nil`.

I'm planning to [drop support for Ruby v2.1](https://github.com/freerange/mocha/issues/628) at which point, this configuration option will be moot.

floehopper:
From [the docs][1]:

> This option only works for Ruby < v2.2.0. In later versions of Ruby
> `nil` is frozen and so a `StubbingError` will be raised if you
> attempt to stub a method on `nil`.

I'm planning to [drop support for Ruby v2.1][2] at which point, this configuration option will be moot.

Closes #693.

[1]: https://mocha.jamesmead.org/Mocha/Configuration.html#stubbing_method_on_nil=-instance_method
[2]: https://github.com/freerange/mocha/issues/628

### DIFF
--- a/lib/mocha/configuration.rb
+++ b/lib/mocha/configuration.rb
@@ -1,4 +1,5 @@
 require 'mocha/ruby_version'
+require 'mocha/deprecation'
 
 module Mocha
   # Allows setting of configuration options. See {Configuration} for the available options.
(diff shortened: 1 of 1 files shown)


floehopper:
Released in v2.7.1

PR #699, opened 2024-12-19
Introduce Expectation#thrice

andrewn617:
I was surprised to get a `NoMethodError: undefined method 'thrice' for an instance of Mocha::Expectation` when I tried to do `foo.expects(:bar).thrice`

This PR introduces a `thrice` method which verifies the expected method is called exactly 3 times.


floehopper:
Superseded by #701. Closing.

floehopper:
@andrewn617 Many thanks for your contribution - can you check that #701 meets your expectations? It has been merged into `main`, but not yet released.

PR #737, opened 2025-01-30
Remove workaround for JRuby jar-dependencies issue

floehopper:
A fix for [this issue][1] was released in [JRuby 9.4.11.0][2] and so [this workaround][3] can be removed, since the `Gemfile` is only used for development.

[1]: https://github.com/jruby/jruby/issues/8488
[2]: https://github.com/jruby/jruby/releases/tag/9.4.11.0
[3]: https://github.com/freerange/mocha/pull/690

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,11 +7,6 @@ group :development do
   gem 'minitest'
   gem 'rake'
 
-  if RUBY_ENGINE == 'jruby'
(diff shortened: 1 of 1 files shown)


floehopper:
It looks like another change might have had the side effect of fixing this in JRuby 9.4.10.0 which is what the CircleCI build seems to be using for `jruby-latest`, but this change should still be OK.

Issue #734, opened 2025-01-24
Investigate removing use of `ruby2_keywords`


floehopper:
This was always intended to be a temporary bridge between Ruby v2 and Ruby v3. Maybe we can remove its use; at least when the current Ruby version if > v3?

See also [this issue](https://github.com/freerange/mocha/issues/594).

Earlopain:
It's getting a bit more pressing since ruby plans to deprecate it now https://bugs.ruby-lang.org/issues/22205 (still a few years out though)

It's complicated since mocha still supports old rubies. I think you'd basically need a copy of `ParametersMatcher` and everything around it that's aware of kw params and use that on on 3.0+

floehopper:
@Earlopain 

Thanks for the heads-up. Since I regard Mocha as approximately feature-complete, I wouldn't mind removing support for older versions of Ruby, since in extremis older versions of the Mocha gem can always be used for projects that can't upgrade the Ruby version.

floehopper:
Reading the comments on https://bugs.ruby-lang.org/issues/22205, it looks like https://github.com/rack/rack/pull/2499 could be a useful reference.

Earlopain:
Dropping old ruby versions would make it easier, agreed. For mocha, it should be 3.0 as the minimum though, because there are differences in keyword handling between 2.7/3.0 (unless you want to deal with that). rack gets away with it becasue they just do simple delegation.

mocha does it differently and actually needs to record positional/keyword arguments separately. In that regard some of the PRs from rails like https://github.com/rails/rails/pull/58238 and https://github.com/rails/rails/pull/58420 are better references.

Issue #594, opened 2022-12-13
Expectations with String keys misidentified as keyword arguments


seandilda:
I'm testing mocha 2.0.2 on ruby 3.1.3 and ran into an issue where a parameter matcher specifying a hash with String keys is being treated as keyword arguments by mocha.

Example code:

```ruby
require 'mocha/api'

include Mocha::API
mocha_setup

class Example
  def foo(args); end
end

example = Example.new
example.expects(:foo).with('key' => 'value')
example.foo({ 'key' => 'value' })
```

This produces the following deprecation warning:

```
Mocha deprecation warning at mocha_test.rb:12:in `<main>': Expectation defined at mocha_test.rb:11:in `<main>' expected keyword arguments ("key" => "value"), but received positional hash ({"key" => "value"}). These will stop matching when strict keyword argument matching is enabled. See the documentation for Mocha::Configuration#strict_keyword_argument_matching=.
```

It looks like mocha treated `.with('key' => 'value')` as specifying a keyword argument instead of a Hash argument.   Since keyword argument syntax looks like a hash with symbol keys, it should be safe to assume that any the parameter is intended to be a Hash if any of the keys aren't Symbols.

floehopper:
@seandilda

Thanks for opening this issue. Can you clarify things a bit further...

1. Are you saying that you don't think a deprecation warning should be displayed in this case?

2. Or are you saying that you want the matching to work in both examples even when strict keyword matching is enabled?

Or both 1 & 2?

floehopper:
@seandilda 

Am I right in thinking this is just a special case of #593?

seandilda:
@floehopper  they are definitely related.   I see #593 as an request for enhancement, as I believe the code is likely working as intended, but maybe not the way that would be most useful for users like myself.

I see this issue as a bug as an expectation that is unambiguously a Hash expectation is being misidentified as being a keyword expectation.

floehopper:
@seandilda Thanks for explaining. While I think I see the distinction you are drawing from a user's point of view, in implementation terms I don't think there's going to be any meaningful difference between this and #593. I may have missed something, but if #593 is fixed then that should also fix this issue - is that correct?

seandilda:
There are some slight differences in the issues, but it is possible a general fix for #593 will also fix this issue.

This issue is about how the expected parameters are parsed.  #593 is about how the arguments used when calling the mock are parsed/compared to expected parameters.

floehopper:
[...]

PR #739, opened 2025-02-16
Simplify parameter matcher for hashes

nitishr:
Improves the handling of keyword and positional hash argument matching. Simplifies the logic around matching keyword and positional hashes, removes redundant code, and ensures that the behavior aligns with Ruby's keyword argument handling.

### Key Changes:
1. **Simplified `to_matcher` Method:**
   - Removed the `last` parameter from the `to_matcher` method in `instance_methods.rb`. This simplifies the logic and removes unnecessary complexity in determining whether a matcher is the last in a sequence.

2. **Refactored `PositionalOrKeywordHash` Class:**
   - Removed the `last_expected_value` parameter and related logic from the `PositionalOrKeywordHash` class. The class now focuses on matching hash entries exactly and handling keyword arguments more straightforwardly.
   - Simplified the `matches?` method to handle both positional and keyword hashes without needing to check if the matcher is the last one.
   - Removed redundant methods like `matches_last_actual_value?` and `last_expected_value_is_positional_hash?`.

3. **Updated Tests:**
   - Removed or updated tests that were checking behavior related to the `last` parameter, as it is no longer relevant.
   - Simplified test cases to focus on the core functionality of matching keyword and positional hashes.
   - Updated the `strict_keyword_argument_matching_test.rb` to reflect the new behavior, ensuring that positional and keyword arguments are matched correctly.


nitishr:
@floehopper I started some [comprehension refactoring](https://martinfowler.com/articles/workflowsOfRefactoring/#comprehension) in order to investigating #594 and #734. In the process, I noticed a number of opportunities for simplification, and the PR is the result.

PR #741, opened 2025-02-18
Builds fail fast on any ruby warnings

### LINKED ISSUE #729: Consider failing fast on any Ruby warnings
I've lost count of the number of times I've accidentally introduced code that causes a warning and only later noticed it.

We can probably use a similar approach as in `ObjectInspectTest#ignoring_warning`.

nitishr:
Closes #729 

Introduces and uses a reusable command in CircleCI config for failing fast on warnings. Addresses resulting warnings that originally caused builds to fail with the updated configuration.
# Changes:
1. **CircleCI Configuration:**
  - Introduce a reusable run-fail-fast command to execute commands and fail on any warnings in the output.
  - Conditionally set RUBYOPT for all Ruby versions except ruby:2.2 (as that version doesn't support the frozen-string-literal flag).
2. **Update Gemfile:** Add the rdoc gem to the development group (as rdoc will no longer be part of the default gems starting from Ruby 3.5.0)
3. **Fix Resulting Warnings:**
  - Update lib/mocha/deprecation.rb to use @logger ||= Logger.new to silence uninitialized ivar warning on older ruby versions.
  - Add a JRuby-specific warning filter in lib/mocha/object_methods.rb to suppress unnecessary warnings.
  - Modify test/acceptance/loose_keyword_argument_matching_test.rb to remove unused variable.

# Benefits:
- **Improved Maintainability:** The CI configuration is more maintainable and easier to update.
- **Faster Feedback Loop:** The run-fail-fast command ensures that the CI pipeline fails immediately if any warnings are encountered during test execution, providing faster feedback.

# Testing:
The CI pipeline has been tested to ensure that it correctly fails on warnings for the specified Ruby versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,6 @@ jobs:
         type: string
     docker:
       - image: << parameters.docker-image >>
-    environment:
-      RUBYOPT: --enable-frozen-string-literal
     steps:
       - checkout
       - run: ruby --version
@@ -23,6 +21,10 @@ jobs:
       - run: gem --version
       - run: bundle --version
       - run: bundle install --gemfile=<< parameters.gemfile >>
+      - run: |
+          if [ "<< parameters.docker-image >>" > "ruby:2.2" ]; then
+            export RUBYOPT="--enable-frozen-string-literal"
+          fi
       - when:
           condition:
             equal: ["Gemfile", << parameters.gemfile >>]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,18 @@
 version: 2.1
 
+commands:
+  run-fail-fast:
+    description: "Execute command and fail on any warning in output"
+    parameters:
+      command:
+        type: string
+    steps:
+      - run: << parameters.command >> 2>&1 | tee output.log
+      - run: |
+          if grep -q "warning:" output.log; then
+            echo "Warnings found in test output. Failing the job."
+            exit 1
+          fi
 jobs:
   build:
     parameters:
@@ -29,28 +42,32 @@ jobs:
           condition:
             equal: ["Gemfile", << parameters.gemfile >>]
           steps:
-            - run: bundle exec rake test
+            - run-fail-fast:
+                command: bundle exec rake test
       - when:
           condition:
             matches:
               pattern: ".*minitest.*"
               value: << parameters.gemfile >>
           steps:
-            - run: MOCHA_RUN_INTEGRATION_TESTS=minitest bundle exec --gemfile=<< parameters.gemfile >> rake test
+            - run-fail-fast:
+                command: MOCHA_RUN_INTEGRATION_TESTS=minitest bundle exec --gemfile=<< parameters.gemfile >> rake test
       - when:
           condition:
             matches:
               pattern: ".*test-unit.*"
               value: << parameters.gemfile >>
           steps:
-            - run: MOCHA_RUN_INTEGRATION_TESTS=test-unit bundle exec --gemfile=<< parameters.gemfile >> rake test
+            - run-fail-fast:
+                command: MOCHA_RUN_INTEGRATION_TESTS=test-unit bundle exec --gemfile=<< parameters.gemfile >> rake test
       - when:
           condition:
             and:
               - equal: ["ruby:latest", << parameters.docker-image >>]
               - equal: ["Gemfile", << parameters.gemfile >>]
           steps:
-            - run: bundle exec rake test:performance
+            - run-fail-fast:
+                command: bundle exec rake test:performance
             - run: MOCHA_GENERATE_DOCS=1 bundle install --gemfile=<< parameters.gemfile >>
             - run: RUBYOPT=--disable-frozen-string-literal MOCHA_GENERATE_DOCS=1 rake docs docs:coverage
   lint:
--- a/lib/mocha/object_methods.rb
+++ b/lib/mocha/object_methods.rb
@@ -4,26 +4,21 @@ require 'mocha/mockery'
 require 'mocha/instance_method'
 require 'mocha/argument_iterator'
 require 'mocha/expectation_error_factory'
+require 'mocha/ignoring_warning'
 
 module Mocha
   # Methods added to all objects to allow mocking and stubbing on real (i.e. non-mock) objects.
   #
   # Both {#expects} and {#stubs} return an {Expectation} which can be further modified by methods on {Expectation}.
   module ObjectMethods
-    if RUBY_ENGINE == 'jruby'
-      require 'mocha/ignoring_warning'
-      extend IgnoringWarning
+    extend IgnoringWarning
 
-      # @private
-      JRUBY_ALIAS_SPECIAL_METHODS_WARNING = /accesses caller method's state and should not be aliased/.freeze
+    # @private
+    JRUBY_ALIAS_SPECIAL_METHODS_WARNING = /accesses caller method's state and should not be aliased/.freeze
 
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -36,6 +34,10 @@ jobs:
       - run: gem --version
       - run: bundle --version
       - run: bundle install --gemfile=<< parameters.gemfile >>
+      - run: |
+          if [ "<< parameters.docker-image >>" != "ruby:2.2" ]; then
```
floehopper, line 38: If it works I think it might be slightly more expressive to use `> "ruby:2.2"` instead of `!= "ruby:2.2"`.

Gemfile
```diff
@@ -7,6 +7,7 @@ group :development do
   gem 'introspection', '~> 0.0.1'
   gem 'minitest'
   gem 'rake'
+  gem 'rdoc'
```
floehopper, line 10: Unless I'm mistaken, this change is orthogonal to the other changes in this commit. If so, can you split it out into a separate explanatory commit?

lib/mocha/object_methods.rb
```diff
@@ -10,6 +10,18 @@ module Mocha
...
+    if RUBY_PLATFORM == 'java'
+      # @private
+      module JRubyAliasMethodWarningFilter
+        def warn(message, *args)
+          super unless /ObjectMethods#method accesses caller method's state and should not be aliased/.match?(message)
+        end
+      end
+
+      # using public_send as calling extend directly causes YARD to warn about Undocumentable mixin
+      Warning.public_send(:extend, JRubyAliasMethodWarningFilter) # rubocop:disable Lint/SendWithMixinArgument, Style/SendWithLiteralMethodName
```
floehopper, line 22: This looks a bit ugly. Could we avoid the `public_send` workaround (and associated comment) and the disabling of two cops by using an approach more like [this](https://github.com/freerange/mocha/blob/d07e12e192ea5a2b9694a35b429308d12dba0a1b/test/unit/object_inspect_test.rb#L68-L78)?

As an aside, I wonder if it might separately be worth trying to extract this pattern somewhere. However, if so, I'd prefer to do that in a separate PR.

floehopper:
> Closes #729
> 
> Introduces and uses a reusable command in CircleCI config for failing fast on warnings. Addresses resulting warnings that originally caused builds to fail with the updated configuration.

Thank you. I'm broadly in favour of this. I've added some comments below and in some review comments.

I've fixed the formatting issues you fixed in `.circleci/config.yml`, so can you rebase against `main` to remove them from the diff?

While I understand why the commits are in the order they are, I think I'd prefer the commits with the warning fixes to come *before* the changes to the CI build which catches them, i.e. so all the tests pass in each commit.

> # Changes:
> 
>     1. **CircleCI Configuration:**
> 
> 
>     * Introduce a reusable run-fail-fast command to execute commands and fail on any warnings in the output.

This seems like quite a fragile way to detect warnings. Is there definitely no better way to do this. I wonder if it's worth looking around at what other projects do? I haven't thought this through, but could we *globally* patch [`Warning#warn`](https://ruby-doc.org/3.4.1/Warning.html) to raise an exception? Although I'm not sure when `Warning#warn` was added or whether it captures all warnings...

>     * Conditionally set RUBYOPT for all Ruby versions except ruby:2.2 (as that version doesn't support the frozen-string-literal flag).
> 
> 
>     2. **Update Gemfile:** Add the rdoc gem to the development group (as rdoc will no longer be part of the default gems starting from Ruby 3.5.0)

Does the absence of this gem actually trigger a warning at the moment?

> 
>     3. **Fix Resulting Warnings:**
> 
> 
>     * Update lib/mocha/deprecation.rb to use @logger ||= Logger.new to silence uninitialized ivar warning on older ruby versions.
> 
>     * Add a JRuby-specific warning filter in lib/mocha/object_methods.rb to suppress unnecessary warnings.
> 
>     * Modify test/acceptance/loose_keyword_argument_matching_test.rb to remove unused variable.
> 
> 
> # Benefits:
> 
>     * **Improved Maintainability:** The CI configuration is more maintainable and easier to update.
> 
>     * **Faster Feedback Loop:** The run-fail-fast command ensures that the CI pipeline fails immediately if any warnings are encountered during test execution, providing faster feedback.
> 
> 
> # Testing:
> 
> The CI pipeline has been tested to ensure that it correctly fails on warnings for the specified Ruby versions.

Thanks for being so thorough with the testing!

floehopper (changes requested):
See review comments and comment on PR description.

nitishr:
>> While I understand why the commits are in the order they are, I think I'd prefer the commits with the warning fixes to come before the changes to the CI build which catches them, i.e. so all the tests pass in each commit.
>

Good point. Done.

> > ```
> > 1. **CircleCI Configuration:**
> > 
> > 
> > * Introduce a reusable run-fail-fast command to execute commands and fail on any warnings in the output.
> > ```
> 
> This seems like quite a fragile way to detect warnings. Is there definitely no better way to do this. I wonder if it's worth looking around at what other projects do? I haven't thought this through, but could we _globally_ patch [`Warning#warn`](https://ruby-doc.org/3.4.1/Warning.html) to raise an exception? Although I'm not sure when `Warning#warn` was added or whether it captures all warnings...
> 

Rspec seems to take a similar approach to [check_documentation_coverage](https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage)

> > ```
> > 2. **Update Gemfile:** Add the rdoc gem to the development group (as rdoc will no longer be part of the default gems starting from Ruby 3.5.0)
> > ```
> 
> Does the absence of this gem actually trigger a warning at the moment?
> 

Yes, it does - but only when generating docs, so now I only add the gem `if ENV['MOCHA_GENERATE_DOCS']`

> > ```
> > 3. **Fix Resulting Warnings:**
> > 
> > 
> > * Update lib/mocha/deprecation.rb to use @logger ||= Logger.new to silence uninitialized ivar warning on older ruby versions.
> > 
> > * Add a JRuby-specific warning filter in lib/mocha/object_methods.rb to suppress unnecessary warnings.
> > 
> > * Modify test/acceptance/loose_keyword_argument_matching_test.rb to remove unused variable.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > # Benefits:
> > ```
> > * **Improved Maintainability:** The CI configuration is more maintainable and easier to update.
> > 
> > * **Faster Feedback Loop:** The run-fail-fast command ensures that the CI pipeline fails immediately if any warnings are encountered during test execution, providing faster feedback.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > # Testing:
> > The CI pipeline has been tested to ensure that it correctly fails on warnings for the specified Ruby versions.
> 
> Thanks for being so thorough with the testing!

Gemfile
floehopper: Addressed in #747.

.circleci/config.yml
floehopper: Addressed in https://github.com/freerange/mocha/pull/741/commits/c37c574b0ec13cab77e19367e0bcedb54a531b92

lib/mocha/object_methods.rb
floehopper: Addressed in https://github.com/freerange/mocha/pull/741/commits/b578fdcbf134260dba82e6cec4826b310761e8b7 & https://github.com/freerange/mocha/pull/741/commits/3dad2198034d058df236bce90e6dd860f9344bc7.

floehopper (approved):
This all looks great to me - thanks for addressing my comments.

PR #743, opened 2025-02-21
Enhance documentation & enforce 100% coverage

### LINKED ISSUE #707: Consider adding some kind of documentation coverage check
c.f. https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage

nitishr:
Closes #707.

Several improvements focused on documentation:

- A new Rake task `docs:coverage` extracts the documented percentage from yard stats and fails the build if it is below the specified level (currently set to 100%).
- CircleCI configuration runs the new docs:coverage task after generating YARD docs. Disables frozen string literals for the documentation generation step to prevent YARD throwing FrozenError.
- Additional source files (`lib/mocha.rb` and `lib/mocha/integration.rb`) are part of the generated documentation.
- Inline documentation in `lib/mocha.rb` provides an overview of the API, the mocking methods, and integration hooks.
- The `Integration` module and the `Adapter`s under it include module-level comments to clarify their purpose.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
           steps:
             - run: bundle exec rake test:performance
             - run: MOCHA_GENERATE_DOCS=1 bundle install --gemfile=<< parameters.gemfile >>
-            - run: MOCHA_GENERATE_DOCS=1 rake yardoc
+            - run: RUBYOPT=--disable-frozen-string-literal MOCHA_GENERATE_DOCS=1 rake yardoc docs:coverage
   lint:
     docker:
       - image: ruby:3.3
--- a/.yardopts
+++ b/.yardopts
@@ -1,5 +1,6 @@
 --output-dir docs
 --no-private
+lib/mocha.rb
 lib/mocha/api.rb
 lib/mocha/hooks.rb
 lib/mocha/mock.rb
@@ -15,6 +16,7 @@ lib/mocha/expectation_error_factory.rb
 lib/mocha/expectation_error.rb
 lib/mocha/stubbing_error.rb
 lib/mocha/unexpected_invocation.rb
+lib/mocha/integration.rb
 lib/mocha/integration/test_unit/adapter.rb
 lib/mocha/integration/minitest/adapter.rb
 -
--- a/Rakefile
+++ b/Rakefile
@@ -161,7 +161,7 @@ if ENV['MOCHA_GENERATE_DOCS']
       end
 
       covered_percent = match[1].to_f
-      min_coverage = 95.0
+      min_coverage = 100.0
 
       if covered_percent < min_coverage
         puts "Documentation coverage is #{covered_percent}%, which is below the required #{min_coverage}%."
--- a/Rakefile
+++ b/Rakefile
@@ -146,6 +146,31 @@ if ENV['MOCHA_GENERATE_DOCS']
 
   desc 'Generate documentation'
   task 'generate_docs' => %w[clobber_yardoc yardoc checkout_docs_cname checkout_docs_js]
+
+  namespace :docs do
+    desc 'Check documentation coverage'
+    task :coverage do
+      stats_output = `yard stats --list-undoc`
+      puts stats_output
+
+      # Extract the documented percentage from the output
+      match = stats_output.match(/(\d+\.\d+)% documented/)
+      if match.nil?
+        puts 'Error: Could not determine documentation coverage.'
+        exit 1
+      end
+
+      covered_percent = match[1].to_f
+      min_coverage = 95.0
+
+      if covered_percent < min_coverage
+        puts "Documentation coverage is #{covered_percent}%, which is below the required #{min_coverage}%."
+        exit 1
+      else
+        puts "Documentation coverage is #{covered_percent}%, which is above the required #{min_coverage}%."
+      end
+    end
+  end
 end
 
 task 'release' => ['default', 'rubygems:release']
--- a/lib/mocha.rb
+++ b/lib/mocha.rb
@@ -2,5 +2,20 @@
 
 require 'mocha/version'
 
+# Mocha's top level namespace, which also provides the ability to {.configure configure} Mocha's behavior.
+#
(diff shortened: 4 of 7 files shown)


floehopper:
Thanks again for all your work! See my comments inline below.

> Closes #707.
> 
> Several improvements focused on documentation:
> 
>     * A new Rake task `docs:coverage` extracts the documented percentage from yard stats and fails the build if it is below the specified level (currently set to 100%).

The "Extend, Extend, Extend!" section of the [YARD docs](https://yardoc.org/) now mentions [yardstick](https://github.com/dkubb/yardstick) for checking documentation quality & coverag, so perhaps we can use that to simplify this code?

The introduction of a `docs` namespace makes sense, but perhaps we should move the existing doc-related tasks into that namespace. Also it might be time to extract some separate `*.rake` tasks...? This isn't essential for this PR - it could be done in a follow-up PR.

> 
>     * CircleCI configuration runs the new docs:coverage task after generating YARD docs. Disables frozen string literals for the documentation generation step to prevent YARD throwing FrozenError.

We might need to remember to remove the disabling of frozen string literals when #741 is merged.
 
>     * Additional source files (`lib/mocha.rb` and `lib/mocha/integration.rb`) are part of the generated documentation.
> 
>     * Inline documentation in `lib/mocha.rb` provides an overview of the API, the mocking methods, and integration hooks.
> 
>     * The `Integration` module and the `Adapter`s under it include module-level comments to clarify their purpose.

All this new documentation looks great - thank you! ❤️

floehopper:
Also I'd like to be sure we fail if [there are warnings or errors when *generating* docs](https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage#L27-L39).

nitishr:
> Thanks again for all your work! See my comments inline below.
> 
> > Closes #707.
> > Several improvements focused on documentation:
> > ```
> > * A new Rake task `docs:coverage` extracts the documented percentage from yard stats and fails the build if it is below the specified level (currently set to 100%).
> > ```
> 
> The "Extend, Extend, Extend!" section of the [YARD docs](https://yardoc.org/) now mentions [yardstick](https://github.com/dkubb/yardstick) for checking documentation quality & coverag, so perhaps we can use that to simplify this code?
> 

I had considered using Yardstick and chose not to because it seems to be an abandoned project - the last commit was in 2015, and the last comment from a maintainer on any of the 15 opens issues was in 2014. Its only documentation seems to be a pretty basic README.

OTOH, it seems to [have some kind of rules](https://github.com/dkubb/yardstick?tab=readme-ov-file#4-configuration) which _seem_ to be qualitative rather than just coverage checks. So it might be worth considering despite the above concern if we wanna use those additional features. However, that seems out of scope for this task.

> The introduction of a `docs` namespace makes sense, but perhaps we should move the existing doc-related tasks into that namespace. Also it might be time to extract some separate `*.rake` tasks...? This isn't essential for this PR - it could be done in a follow-up PR.
> 

Yes, that was the idea behind introducting the namespace. And I wanted to contain the scope of this PR.

> > ```
> > * CircleCI configuration runs the new docs:coverage task after generating YARD docs. Disables frozen string literals for the documentation generation step to prevent YARD throwing FrozenError.
> > ```
> 
> We might need to remember to remove the disabling of frozen string literals when #741 is merged.
> 

Perhaps I'm not thinking clearly about this but I don't think so. I disabled frozen string literals for the yard step because for some reason YARD throws an error as its internal code does some string concat. I'm not sure why this didn't happen before - perhaps it has to do with the latest ruby version (which is the only version we run that step for.

So, yes, we would want to remove the disabling _at some point_, and that seems unrelated to #741.

> > ```
> > * Additional source files (`lib/mocha.rb` and `lib/mocha/integration.rb`) are part of the generated documentation.
> > 
> > * Inline documentation in `lib/mocha.rb` provides an overview of the API, the mocking methods, and integration hooks.
> > 
> > * The `Integration` module and the `Adapter`s under it include module-level comments to clarify their purpose.
> > ```
> 
> All this new documentation looks great - thank you! ❤️

nitishr:
> Also I'd like to be sure we fail if [there are warnings or errors when _generating_ docs](https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage#L27-L39).

Perhaps part of, or a follow up to #741?

nitishr:
> > > The "Extend, Extend, Extend!" section of the [YARD docs](https://yardoc.org/) now mentions [yardstick](https://github.com/dkubb/yardstick) for checking documentation quality & coverag, so perhaps we can use that to simplify this code?
> > 
> > 
> > I had considered using Yardstick and chose not to because it seems to be an abandoned project - the last commit was in 2015, and the last comment from a maintainer on any of the 15 opens issues was in 2014. Its only documentation seems to be a pretty basic README.
> > OTOH, it seems to [have some kind of rules](https://github.com/dkubb/yardstick?tab=readme-ov-file#4-configuration) which seem to be qualitative rather than just coverage checks. So it might be worth considering despite the above concern if we wanna use those additional features. However, that seems out of scope for this task.
> 
> Ah, OK. That's all very helpful to know. And in the light of that, I'm happy to go with your current approach.

Cool.

And are you OK with this being in a different PR - either #741 or a follow up: fail if [there are warnings or errors when generating docs](https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage#L27-L39)?

Also, for grouping other docs tasks under the docs namespace, I'm happy to do that as part of either this PR or a separate one. Could you let me know your preference?

floehopper:
@nitishr Sorry - I started work on https://github.com/freerange/mocha/pull/745 before I saw this.

floehopper:
> Also I'd like to be sure we fail if [there are warnings or errors when generating docs](https://github.com/rspec/rspec/blob/60a7a65e195953196fb8d1836257909c56d2da85/script/check_documentation_coverage#L27-L39).

I've captured this idea in https://github.com/freerange/mocha/issues/748.

Issue #754, opened 2025-04-20
Regression in `Hash.ruby2_keywords_hash?` behaviour in JRuby v10


floehopper:
I've seen a change in behaviour of `Hash.ruby2_keywords_hash?` in JRuby v10 manifested as some test failures in [this Mocha CI build](https://app.circleci.com/pipelines/github/freerange/mocha/857/workflows/99ea71bc-a02f-4f35-aeac-6e61b2c97a33/jobs/12856).

Reported to JRuby in https://github.com/jruby/jruby/issues/8784.

floehopper:
Fixed in https://github.com/jruby/jruby/pull/8785. Now we have to wait for a JRuby release that includes it!

floehopper:
Mysteriously this seems to have been fixed in a new version (!) of [JRuby v10.0.1.0](https://github.com/jruby/jruby/releases/tag/10.0.1.0), i.e the SHA seems to have changed from `79cf1e4aec` -> `0f10d1dfdf`:

```
ruby --version
jruby 10.0.0.1 (3.4.2) 2025-05-07 79cf1e4aec OpenJDK 64-Bit Server VM 21.0.7+6-LTS on 21.0.7+6-LTS +indy +jit [x86_64-linux]
```

```
ruby --version
jruby 10.0.1.0 (3.4.2) 2025-07-17 0f10d1dfdf OpenJDK 64-Bit Server VM 21.0.7+6-LTS on 21.0.7+6-LTS +indy +jit [x86_64-linux]
```

Closing.

floehopper:
I'm not sure whether it's a typo or I had a dyslexic moment, but the comment above suggests the JRuby version *did* change from v10.0.0.1 -> v10.0.1.0. So perhaps that explains how the problem was fixed.

PR #755, opened 2025-05-09
Fix compatibility with ruby 3.5

Earlopain:
In 3.5, most of the `cgi` gem is removed. Only methods relating to escaping/unescaping are retained.

That means, on 3.5 you get this error:
> NoMethodError: undefined method 'parse' for class CGI

This replaces the cgi usage with https://docs.ruby-lang.org/en/master/URI.html#method-c-decode_www_form, that also returns this information.

It's in array form, so it first needs to be converted to an hash.

This method is available on all versions going back to at least Ruby 2.0 so compatibility should be ok with this.

https://bugs.ruby-lang.org/issues/21258

### DIFF
--- a/lib/mocha/parameter_matchers/equivalent_uri.rb
+++ b/lib/mocha/parameter_matchers/equivalent_uri.rb
@@ -2,7 +2,6 @@
 
 require 'mocha/parameter_matchers/base'
 require 'uri'
-require 'cgi'
 
 module Mocha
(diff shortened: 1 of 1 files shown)


Earlopain:
JRuby fails because of https://github.com/freerange/mocha/issues/754

floehopper:
@Earlopain Thanks for contributing this 👍

PR #757, opened 2025-06-25
Add missing quote to docstrings.

bensherman:
I was reading the docs for parameter matching in tests, and found there is a missing quote at the end of a url in the docstrings for the equivalent_uri matcher.

❤️ Mocha. Thanks.


floehopper:
@bensherman Thanks for taking the time to fix it - much appreciated!

The JRuby build failure is [a known issue](https://github.com/freerange/mocha/issues/754), so I'm happy to merge this.

PR #789, opened 2026-02-12
Add a `:warn_with_location` option to stubbing non-existent methods

issyl0:
- Display a warning message that includes the file and line number where the stub was made, making it easier to locate problematic stubs in test code when you're running a lot of tests at once right in front of you, or trying to search test run logs.
- I considered adding the file path to the existing `:warn` option, but that might break others' workflows?
- Is this _too_ verbose?

Before:

```
WARNING: stubbing non-existent method: #<#<Class:0x00007f2a9dc4e4c8>:0x2e9810>.user_ids
WARNING: stubbing non-existent method: #<#<Class:0x00007f36c3d47c80>:0x2e97c0>.repo_ids
```

After:

```
WARNING: stubbing non-existent method: #<#<Class:0x00007d4c1d557c60>:0x64140>.user_ids at foo_test:59:in 'block (2 levels) in <class:ViewTest>'
WARNING: stubbing non-existent method: #<#<Class:0x00007d4c1d557c60>:0x66e10>.repo_ids at foo_test.rb:38:in 'block (2 levels) in <class:ViewTest>'
```

### DIFF
--- a/lib/mocha/configuration.rb
+++ b/lib/mocha/configuration.rb
@@ -134,9 +134,10 @@ module Mocha
     #
     # When +value+ is +:allow+, do nothing. This is the default.
     # When +value+ is +:warn+, display a warning.
+    # When +value+ is +:warn_with_location+, display a warning including file and line number where the stub was made.
     # When +value+ is +:prevent+, raise a {StubbingError}.
     #
-    # @param [Symbol] value one of +:allow+, +:warn+, +:prevent+.
+    # @param [Symbol] value one of +:allow+, +:warn+, +:warn_with_location+, +:prevent+.
     #
     # @example Preventing stubbing of a non-existent method
     #
--- a/lib/mocha/logger.rb
+++ b/lib/mocha/logger.rb
@@ -6,8 +6,10 @@ module Mocha
       @io = io
     end
 
-    def warn(message)
-      @io.puts "WARNING: #{message}"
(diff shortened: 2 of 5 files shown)


issyl0:
Entirely my bad for not copying the whole thing into the PR! And no rush on this at all, don't interrupt your leave!

floehopper:
@issyl0

I've started looking into this and I agree that displaying the source location for Mocha warnings would be a sensible change.

However, I'd like to apply it to all of the warning types. Looking at the existing code has highlighted a couple of other things that I'd like to address at the same time:
* Mocha's deprecation warning mechanism already has [something like this](https://github.com/freerange/mocha/blob/50844e9acb37e9fd86c767d73f6c79b534fb01c0/lib/mocha/deprecation.rb#L8-L12)
* There's some inconsistency (but also some effective duplication) between how Mocha is reporting warnings vs deprecation warnings.

I've made a start on addressing all this myself. I'm making some significant changes to the structure of the code, so while I plan to address the issue you've raised, I'm probably not going to actually use your changes directly. I hope that's OK! I really appreciate that you raised the issue though. To reflect this, I've created a [new issue](https://github.com/freerange/mocha/issues/792) and I'm going to close this PR.

As an aside, it also made me spot [this bug which I fixed in v3.0.2](https://github.com/freerange/mocha/pull/790) - so thank you for that too!

PR #794, opened 2026-02-15
Format warnings consistently and include source location

### LINKED ISSUE #792: Make deprecation warnings & warnings more consistent and include source location for both
@issyl0 opened https://github.com/freerange/mocha/pull/789 motivated by wanting to see the source location for lots of Mocha warnings in the build output for a project without the noise of full stack traces that would be generated using the `:prevent` configuration mode. This makes a lot of sense to me and I'd like to implement it for all Mocha warning types; not just for stubbing non-existent methods.

When I started investigating I noticed:
* Mocha's deprecation warning mechanism already [includes the source location](https://github.com/freerange/mocha/blob/50844e9acb37e9fd86c767d73f6c79b534fb01c0/lib/mocha/deprecation.rb#L8-L12).
* However, there's some inconsistency (but also some effective duplication) between how Mocha is reporting warnings vs deprecation warnings.

So I've opened this issue in place of #789.

floehopper:
This changes both deprecation and non-deprecation warnings to share the same "logger" which avoids some duplication and makes it easier to format all warning messages consistently and including the source location where the problem was triggered. These changes were prompted by #789 and addresses #792. It introduces some changes to the formatting of warning messages:

### Deprecation warnings

* Before change: `WARNING at <location>: <message>`
* After change: `Mocha deprecation warning: <message> (at <location>)`

### Other warnings

* Before change: `WARNING: <message>`
* After change: `Mocha warning: <message> (at <location>)`

Note that I've previously made it clear in the `README` that [the documented public API does not include the content or format of messsages displayed to the user][1].

I've also removed the source code element of the caller location included in the deprecation warnings, i.e. only the the source file path and the line number are included. This makes the warning messages less verbose without losing much useful information and is consistent with how Ruby's own deprecation warnings are displayed. This addresses [this feedback](https://github.com/freerange/mocha/issues/792#issuecomment-4005188814) from @issyl0.

Closes #792.

[1]: https://mocha.jamesmead.org/index.html#semantic-versioning


floehopper:
I tested this in action by running the following Ruby code:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'minitest'
  gem 'mocha'
end

require 'minitest/autorun'
require 'mocha/minitest'
require 'mocha/deprecation'

  def test_foo
    Mocha.configure do |c|
      c.stubbing_non_existent_method = :warn
    end

    foo = Object.new
    foo.stubs(:foo)
  end

  def test_bar
    Mocha.configure do |c|
      c.strict_keyword_argument_matching = false
    end

    mock = mock()
    kwargs = { key: 42 }
    mock.expects(:method).with(**kwargs)

    mock.method({ key: 42 })
  end
end
```

Which generated the following output (as expected):

```
Mocha deprecation warning: Expectation defined at foo.rb:32 expected keyword arguments (key: 42), but received positional hash ({key: 42}). These will stop matching when strict keyword argument matching is enabled. See the documentation for Mocha::Configuration#strict_keyword_argument_matching=. (at foo.rb:34)
.Mocha warning: stubbing non-existent method: #<Object:0x510>.foo (at foo.rb:22)
.

Finished in 0.000640s, 3125.0001 runs/s, 1562.5000 assertions/s.

2 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

Issue #574, opened 2022-10-18
Consider using Kernel#caller_locations vs Kernel#caller for richer a richer backtrace


floehopper:
This seems to have been available since Ruby v2.0 and so, now we've dropped support for earlier versions of Ruby, it should be safe to use it.

* https://www.rubydoc.info/stdlib/core/2.0.0/Kernel:caller_locations
* https://jemma.dev/blog/enumerable-tally

PR #801, opened 2026-06-09
Avoid `Kernel#caller` as much as possible

hmcguire-shopify:
Ref #574

`#on_stubbing` calls `#check` three times without passing a backtrace, so `#caller` is called three times as well. However, the backtrace isn't actually used unless an exception is raised by `#check`, so these (expensive) invocations are completely wasteful.

This commit makes `#caller` lazy, because it currently shows up as one of the top three hottest individual methods in our test suite (and more than 2% of all allocations).

The other changes from `#caller` to `#caller_locations` are less hot, but `#caller_locations` is more efficient than `#caller` so I included them as well.

### DIFF
--- a/lib/mocha/mockery.rb
+++ b/lib/mocha/mockery.rb
@@ -150,13 +150,13 @@ module Mocha
 
     private
 
-    def check(action, description, signature_proc, backtrace = caller)
+    def check(action, description, signature_proc, backtrace = nil)
       treatment = Mocha.configuration.send(action)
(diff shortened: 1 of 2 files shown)


floehopper:
@hmcguire-shopify Thanks for opening this PR. I've tweaked a few things and opened #802 in its place, so I'm going to close this one.

hmcguire-shopify:
That works for me, thanks!

PR #804, opened 2026-09-21
Remove internal `_method` as an alias of `method`

Earlopain:
Ruby will warn because the module may be included into `BasicObject`, which will not have `method` available.

> bundler/gems/mocha-7826b6b4e29c/lib/mocha/object_methods.rb:20: warning: the fallback to Object for alias of 'method' in module 'Mocha::ObjectMethods' is deprecated and will be removed in Ruby 4.3

https://bugs.ruby-lang.org/issues/22276

As a plus, jruby doesn't warn for this style.


Earlopain:
I guess jruby still needs the warning supression in tests after all🤷

floehopper:
@Earlopain Thanks for submitting this. FYI I've just [fixed](https://github.com/freerange/mocha/pull/805) the Rubocop issues in `main`. The original implementation was a bit ugly, so I'm going to take this opportunity to see whether there's a better solution when I've got a bit of time.

Notes to self:
* Originally introduced in https://github.com/freerange/mocha/pull/77
* Now tested in https://github.com/freerange/mocha/blob/99da677165a5ef21cc5ad21da733e82fc6aa2636/test/unit/object_methods_test.rb#L61-L65

Earlopain:
It's actually dead code. A better solution would perhaps have been to store `Kernel.instance_method(:method)` in a constant  and call it directly where it's used instead of pulling it in the module like this but removing it alltogether is probably cleaner 😅

This is the commit where it stoped being used https://github.com/freerange/mocha/commit/0cb193a303a3a2d2b57a7bba129152fd67b7b413 (or more precisely https://github.com/freerange/mocha/commit/30ac58936999771cd29ed11852dc084ab8053b95)

